PR #10, opened 2023-07-25
Added the interactive learnbyheart page

A-meerdervan:
I created the alexaddlearnbyheart branch.
It now also contains questions on what capital belongs to what country, this is just for Joep to get a feel for what the 
experience is to use this when you do not know any of the answers when starting. How the learning feels. 
The capitals subsets should be removed from the aifacts.txt file before merging with the main branch.

What is added:
A page at /learnbyheart that is an interactive way to learn about AI in general and xrisk and the pause. 
All content is controlled by the aifacts.txt file
I made a google doc with instructions and all of the information that can be learned, to use for updating the aifacts file with a
copy paste operation.
There are tripples of question,answer,context on each line.
The user can select between multiple subsets of questions
The user gets asked the question, learns from seeing the answer and gets the question asked again later to drill in the knowledge. 

List of commit messages:
- Created a page at /learnbyheart that is an interactive web app where the user can learn ai facts by heart through repetition
- minor update
- I added the subsets of capital country pairs, need to be removed before merging


src/routes/learnbyheart/+page.svelte
```diff
@@ -0,0 +1,685 @@
...
+	let dataset = [{question: 'Loading', answer: 'nonsense', weight: -1, context:'Cool context here'}];
+	let datasets = [];
+	let datasetNames = [];
+	let datasetNamesAndIndexes =[];
+
+	// This onMount function is needed to run this loading of a file operation in the browswer, because the server did not allow a 
+	// fetch operation. The text file contains multiple subsets with lines: question;answer;context
+	onMount(async () => {
+	try {
+		const response = await fetch(pathToData);
```
joepio, line 62: This will not work with server-side rendering, you can't use fetch here, needs to be in the load function.

https://kit.svelte.dev/docs/load

src/routes/learnbyheart/+page.svelte
```diff
@@ -0,0 +1,685 @@
...
+	// This onMount function is needed to run this loading of a file operation in the browswer, because the server did not allow a 
+	// fetch operation. The text file contains multiple subsets with lines: question;answer;context
+	onMount(async () => {
+	try {
+		const response = await fetch(pathToData); 
+		const content = await response.text();
+		const lines = content.split('\n');
+		dataset = [];
+		let datasetIndex = -1; // done to make the first datasetIndex start at -1+1 = 0.
+		for (let i = 0; i < lines.length; i++) {
```
joepio, line 67: All this parsing logic isn't needed if you migrate the .txt to something like JSON. You can probably use chatGPT to help with this!

joepio:
Cool stuff! It's working locally :)

Thanks for making this 😄 

Some UX feedback:

- I'd suggest switching to a multiple-choice input. It is more conventional, easy to use and it works on touchscreen devices.
- The question feedback is a little weirdly positioned. I think it would make more sense to show it as a pop-up. Also, show the prev question + answer + feedback in the same screen before moving on to the next question.
- Showing the weight seems like an odd thing. As an end-user, it is confusing
- The next / prev browser buttons do not work. I'd suggest storing the question number in a query param so it becomes part of the browser's history

Code feedback:

- Put the .txt data in a structured format. I'd go for JSON. You don't need custom parsing logic that way, it becomes interoperable, it's auto-highlighted in your IDE and you can do type checking. 

Also made some comments!

Overall, cool work :)

PR #11, opened 2023-09-08
Patch 1

UFO-101:


src/posts/urgency.md
```diff
@@ -30,28 +24,25 @@ It's because GPT-4 has been fine-tuned to not mislead people. It's not good for
...
+- **We like to think that we're special**. Humans like to feel that we are special. If an AI can do what we can do, we're not special anymore. This is a hard pill to swallow, and the [brain has many defense mechanisms to avoid this](psychology-of-x-risk).
 - **We're really bad at exponential growth**. We tend to structurally and predictably underestimate how exponential growth cumulates over time. This has been shown in [scientific studies](https://www.researchgate.net/figure/Underestimation-of-exponential-growth-a-shows-the-participants-prediction-of-the_fig4_351171143).
 
-But there are still some things that an AI can't do yet, luckily.
+Luckily there are still some things that an AI can't do yet.
 It cannot [hack better than the best hackers](/cybersecurity-risks) and it cannot do AI research as well as the best AI researchers.
-**When we reach either of these thresholds, we will be in danger**.
+**When we reach either of these thresholds, we will in a new regime of increased risk**.
```
joepio, line 45: be in?

src/posts/cybersecurity-risks.md
```diff
@@ -27,31 +27,8 @@ or use [WiFi to see humans through walls](https://www.marktechpost.com/2023/02/1
 There will most likely come a point where an AI is better at hacking than the best human hackers.
 This can go wrong in many ways.
 
-- **Nuclear**: Cyberweapons can be used to gain access to nuclear materials, meltdown a reactor, convince humans they need to fire nuclear missiles, or even fire them automatically.
```
joepio, line 30: This deserves a source, but it should not be removed in its entirety. I remember reading some articles about it .

- https://www.nytimes.com/2017/03/14/opinion/why-our-nuclear-weapons-can-be-hacked.html
- https://www.nti.org/analysis/articles/cyber/

Deserves more discussion probably

src/posts/cybersecurity-risks.md
```diff
@@ -27,31 +27,8 @@ or use [WiFi to see humans through walls](https://www.marktechpost.com/2023/02/1
...
-No single person could do this, but a sufficiently capable AI could.
-This type of Super Virus could spread to effectively all machines that have some sort of connection to other computers (through WiFi, Bluetooth, UTP, etc.).
-It could then have full control over these machines.
-It could steal data, spread itself, use the hardware for computations, or simply make the machine unusable.
-
-If creating these types of viruses becomes technically feasible, someone will likely do it eventually.
-It could happen by accident (e.g. by a [misaligned AI](/xrisk)), or it could happen on purpose (e.g. as a cyber weapon by a nation or terrorist).
-And when that happens, the effects could be devastating.
-
-If effectively all computers are disabled, society breaks down.
```
joepio, line 50: This point about societal breakdown and combining multiple zero-days in one virus is important to the main premise of the article: cybersecurity skills can threaten many lives.

It's an extreme scenario, and if there are things wrong in the argument I'd love to hear them

src/posts/faq.md
```diff
@@ -18,15 +18,12 @@ AI safety researchers (who are the experts on this topic) are divided on this qu
 Note that no (surveyed) AI safety researchers believe that there's a 0% chance.
 However, there might be selection bias here: people who work in the AI safety field are likely to do so because they believe preventing bad AI outcomes is important.
 
-If you ask AI researchers in general (not safety specialists), this number drops to a [mean value of around 14%](https://aiimpacts.org/2022-expert-survey-on-progress-in-ai/).
+If you ask AI researchers in general (not safety specialists), this number drops to a [mean value of around 14%](https://aiimpacts.org/2022-expert-survey-on-progress-in-ai/), with a median of 5%.
 A small minority, about 4% of them, believe that the alignment problem is not a real problem.
 Note that there might be a selection bias here in the opposite direction: people who work in AI are likely to do so because they believe AI will be beneficial.
 
 _Imagine you're invited to take a test flight on a new airplane_.
-The airplane safety experts on average think there's a 30% chance it would crash.
```
joepio, line 26: Why remove this line? It makes the analogy more powerful and it's backed up by data.

src/posts/faq.md
```diff
@@ -18,15 +18,12 @@ AI safety researchers (who are the experts on this topic) are divided on this qu
...
-The plane engineers think there's a 14% chance of crashing.
-About 4% of engineers are saying that there's no chance of crashing, it is uncrashable.
-
+The plane engineers think there's a 5% chance of crashing.
```
joepio, line 26: That's not their average estimate - it's 14%. Why use the median value here?

Perhaps if you change the sentence to "the average plane engineer thinks there's a 5% chance of crashing" that makes sense.

Or "The plane engineers, on average, think there's a 14% chance of crashing".

I'd go for using the average instead of the median value.

src/posts/summit.md
```diff
@@ -12,24 +12,14 @@ This needs to happen on an international level.
 
 The primary goal of PauseAI was to convince one government to organize such a summit.
 Just 5 weeks after the first PauseAI protest, the UK government has [announced](https://www.gov.uk/government/news/uk-to-host-first-global-summit-on-artificial-intelligence) that they will host an AI safety summit in on November 1-2 in London.
-This is amazing news, and we are very grateful to the UK government for taking this step.
+This is amazing news, and we are very grateful to the UK government for taking this step (we are not claiming we are solely responsible for this summit :).
```
joepio, line 15: I agree with adding this message (never wanted to insinuate that we have any responsibility, by the way).

Maybe move it to a separate sentence, and one line higher, and drop the `:`

src/posts/summit.md
```diff
@@ -12,24 +12,14 @@ This needs to happen on an international level.
...
 However, this does not mean that we're done.
 
 ## What is a summit?
 
-A summit is a meeting of national governments, hosted by one of them.
```
joepio, line 20: Ok with removing this part. But by removing this, the title of this section and the next paragraph feel odd

src/posts/xrisk.md
```diff
@@ -5,11 +5,9 @@ description: Why AI is a risk for the future of our existence, and why we need t
 
 ## Experts are sounding the alarm
 
-AI researchers on average [believe](https://aiimpacts.org/2022-expert-survey-on-progress-in-ai/) there's a 14% chance that once we build a superintelligent AI, it will lead to "very bad outcomes (e.g. human extinction)".
+AI researchers on average [believe](https://aiimpacts.org/2022-expert-survey-on-progress-in-ai/) there's a 5% chance that once we build a superintelligent AI, it will lead to "very bad outcomes (e.g. human extinction)".
 
-Would you choose to be a passenger on a test flight of a new plane when airplane engineers think there’s a 14% chance that it will crash?
-
-When you ask AI safety researchers (the people who actually research the safety risks of AI), the average estimate for bad outcomes [grows to 30%](https://forum.effectivealtruism.org/posts/8CM9vZ2nnQsWJNsHx/existential-risk-from-ai-survey-results).
```
joepio, line 12: Why leave this out? I think AI safety researchers are far more likely to have given this subject serious thought than an average AI researcher. Their estimates are valuable.

src/posts/xrisk.md
```diff
@@ -5,11 +5,9 @@ description: Why AI is a risk for the future of our existence, and why we need t
 
 ## Experts are sounding the alarm
 
-AI researchers on average [believe](https://aiimpacts.org/2022-expert-survey-on-progress-in-ai/) there's a 14% chance that once we build a superintelligent AI, it will lead to "very bad outcomes (e.g. human extinction)".
+AI researchers on average [believe](https://aiimpacts.org/2022-expert-survey-on-progress-in-ai/) there's a 5% chance that once we build a superintelligent AI, it will lead to "very bad outcomes (e.g. human extinction)".
```
joepio, line 8: The average value is 14%, not 5%.

It's misleading to use the median and say `average`. Why should we mislead people about this?

src/posts/xrisk.md
```diff
@@ -128,33 +121,23 @@ This is an AI, using the aforementioned AutoGPT + GPT-4, that is instructed to "
...
-But what would happen if it was extremely intelligent?
-What could have happened if it was smart enough to [hack into nuclear silos](https://t.co/UNPERx9NdL)?
+Luckily ChaosGPT didn't get very far in its quest for dominance.
+The reason it didn't get very far: it wasn't that smart.
+As language models improve the threat from AIs such as ChaosGPT will increase.
 
-The dangerous capacity is intelligence.
-There is no need for consciousness, a body, or any underlying motivation.
-It does not matter if it is just a chatbot - it can still be an existential threat with the wrong prompt, a tool like AutoGPT and sufficient intelligence.
-If an AI is smart enough to take over, at some point in time it will try to do this, simply because people will try all sorts of things.
```
joepio, line 139: This argument feels too important to leave out. I even think it deservers an entire paragraph, instead of one sentence.

Every time we use an AI, we create a new instance. Every prompt is different, and every prompt means that instance could land in a configuration where it decides that spreading itself is sensible. 

It only has to go wrong once. We can have 99999 AI's that are friendly oracles that answer our questions and do their tasks, and only one instance that decides that it should seek power, and we're fucked.

src/posts/xrisk.md
```diff
@@ -128,33 +121,23 @@ This is an AI, using the aforementioned AutoGPT + GPT-4, that is instructed to "
...
 [Read more about urgency](/urgency).
 
 ## AI companies are locked in a race to the bottom
 
-Building artificial intelligence used to be mostly a research/university thing.
-In just a couple of months, it became a multi-billion dollar industry.
-And the companies involved are rushing to build the most powerful AI they can.
-They are pressured by investors to focus on capabilities, not safety.
-We need to give these AI companies a reason to focus on safety.
+OpenAI, DeepMind and Anthropic want to develop AI safely, but they are forced by various incentives to compete with each other and keep moving faster.
```
joepio, line 141: Not entirely sure about this tone shift. We're still a group of activists, and we should keep pressure on these AI companies. They're not the victims here. It's not that Moloch is actually putting a gun to their heads.

joepio (changes requested):
Wow, thanks for all these changes! Lots of great improvements.

I've left a bunch of comments where I don't (yet) agree with the changes. 

No comment = great stuff! :)

src/posts/summit.md
UFO-101: Changed it a bit

src/posts/summit.md
UFO-101: Changed title and next line

src/posts/xrisk.md
UFO-101: Changed to "The average AI researcher believes".

See discussion above. I think giving conservative estimates is more persuasive, and strategically choosing the highest of the mean or median for each survey is in fact statistical manipulation.

src/posts/xrisk.md
UFO-101: See discussion above. I agree their estimates are more valuable. But only because I _already_ believe in AI x-risk.

src/posts/xrisk.md
```diff
@@ -90,14 +85,12 @@ An AI could have any goal.
 Maybe it wants to calculate pi, maybe it wants to cure cancer, maybe it wants to self-improve.
 This depends on how it is trained.
 But even though we cannot tell what a superintelligence will want to achieve, we can make predictions about its sub-goals.
-We can [mathematically prove](https://arxiv.org/pdf/1912.01683.pdf) what a rational machine will pursue in _virtually any goal_:
```
UFO-101, line 93: Btw, this paper has been dis-endorsed by its own author as irrelevant.

src/posts/xrisk.md
UFO-101: Sure, I think adding a paragraph on this could be good. But I think the way it is currently stated is hard to understand.

src/posts/xrisk.md
UFO-101: Updated this section.

(Discussing investors is complicated because OpenAI and Anthropic have unusual corporate structures to try to mitigate this problem.)

src/posts/faq.md
joepio: The number does not reflect how many people the risk is real, but how large the risk is.

AI safety researchers are the domain experts here.

If you want to know how many people don't believe the alignment problem is real, that one is included too. That would be analogous to saying "the earth is flat"

src/posts/cybersecurity-risks.md
joepio: As soon as the capability to find novel zero days exists, AI can create such a super virus, because AI works extremely fast and can be scaled up to checking all code bases.

I expect this capability to emerge before full on superintelligence.

src/posts/xrisk.md
joepio: Choosing the Lowe number because it's more pursuasive is explicitly manipulative though.


If anything, I'd rather revert to the "half of scientists think there's a larger than 10% chance", which shows that it's not a small number of extremes pushing the number.

src/posts/faq.md
joepio: Good points. On manipulation I disagree, as we're not choosing the highest number, we're simply showing the average - an easier to understand, more commonly used statistical metric than the median. And we're not showing different numbers from different studies.

On pursuasiveness, you may have a point, it might convince more people if we use the lower number. But that _is_ being manipulative. 

I believe we should be honest and speak the truth. We should not choose the lower number with the less commonly used statistical tool to make our case more convincing.

src/posts/xrisk.md
joepio: Didn't know that, looks like I've got some reading to do. 

This post is probably a good start: https://www.lesswrong.com/posts/pdaGN6pQyQarFHXF4/reward-is-not-the-optimization-target

src/posts/cybersecurity-risks.md
UFO-101: Updated

src/posts/faq.md
UFO-101: Manipulating the data to be as charitable as possible to the other side seems fine. I don't agree the average is the default choice here. I think in an opinion survey like this it is probably more standard to report the median.

src/posts/faq.md
```diff
@@ -18,15 +18,12 @@ AI safety researchers (who are the experts on this topic) are divided on this qu
...
+If you ask AI researchers in general (not safety specialists), this number drops to a [mean value of around 14%](https://aiimpacts.org/2022-expert-survey-on-progress-in-ai/), with a median of 5%.
 A small minority, about 4% of them, believe that the alignment problem is not a real problem.
 Note that there might be a selection bias here in the opposite direction: people who work in AI are likely to do so because they believe AI will be beneficial.
 
 _Imagine you're invited to take a test flight on a new airplane_.
-The airplane safety experts on average think there's a 30% chance it would crash.
-The plane engineers think there's a 14% chance of crashing.
-About 4% of engineers are saying that there's no chance of crashing, it is uncrashable.
-
+The average plane engineers thinks there's a 5% chance of crashing.
```
joepio, line 26: Using the median value while saying "average" is misleading. Also, I don't think we should use the median value here - we should use the mean.

Say we had 10 researchers, 4 of which had a p(doom) of 100% and 6 of them 0%. The median p(doom) would be 0%, the mean 40%.

How worried should you be in this case? 40% or 0%? Really, the relevant number here is the _average_, not the median!

The median is misleading.

src/posts/faq.md
joepio: AI safety researchers are the actual experts here, so I don't think the analogy is unfair. If experts are _this_ worried about it, we should be worried, too.

src/posts/xrisk.md
```diff
@@ -5,11 +5,9 @@ description: Why AI is a risk for the future of our existence, and why we need t
 
 ## Experts are sounding the alarm
 
-AI researchers on average [believe](https://aiimpacts.org/2022-expert-survey-on-progress-in-ai/) there's a 14% chance that once we build a superintelligent AI, it will lead to "very bad outcomes (e.g. human extinction)".
+The average AI researcher [believes](https://aiimpacts.org/2022-expert-survey-on-progress-in-ai/) there's a 5% chance that once we build a superintelligent AI, it will lead to "very bad outcomes (e.g. human extinction)".
```
joepio, line 8: Using median + using the word average is misleading (see above)

src/posts/xrisk.md
joepio: Fair

src/posts/faq.md
```diff
@@ -83,7 +80,7 @@ These companies have been very slow to acknowledge x-risk, considering that many
 So in our view, the AI companies are not pushing the x-risk narrative, they have been reactive to others pushing it, and have waited with their response until it was absolutely necessary.
 
 The business incentives point in the other direction: companies would rather not have people worry about the risks of their products.
-Virtually all companies downplay risks to attract customers and investments (remember the Oceangate submarine?), not exaggerate them.
```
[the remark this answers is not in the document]
joepio: Really? I expect most people to have followed this story in some degree. Maybe I should add a link: https://en.wikipedia.org/wiki/Titan_submersible_implosion

joepio (changes requested):
Thanks for the changes again! See comments

src/posts/faq.md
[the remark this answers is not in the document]
UFO-101: Regardless I don't think this is a strong argument. Individual stories can be found to support any position.

src/posts/faq.md
UFO-101: Okay reverted to 14%

src/posts/xrisk.md
UFO-101: Reverted.

src/posts/faq.md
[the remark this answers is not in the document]
joepio: fair point

joepio (approved):

PR #29, opened 2024-01-02
Mark external links by appending an icon

### LINKED ISSUE #28: Mark external links
When browsing the website, I found it frustrating that I didn't know which links I could use to navigate around the website and which ones would take me elsewhere. Maybe it would be a good idea to mark external links.

I will submit a PR that potentially solves the issue by appending an icon, but I don't know if that's the best solution and if it should be implemented differently, so I also opened this issue.

Wituareard:
I implemented a style that potentially closes #28 by appending an icon to all external links, similar to how Wikipedia handles it. Personally I don't find it to be too distracting, but I contribute to Wikipedia, so I am used to the design and therefore that's probably not the best metric.

This is how it makes the text look:
grafik

And this is how it makes the footer look:
grafik

(I hope I didn't spam everyone's emails too much with my failed attempts to fork the repo and use this account locally.)


joepio:
Awesome! Thanks :D

PR #30, opened 2024-01-03
Fix grammar error

Wituareard:
An alternative wording would be "how to do it safely" referring to the process of developing the systems.


joepio:
Great, thanks!

PR #39, opened 2024-01-07
More countercounterarguments

Pato-desu:


Pato-desu:
mixed a bunch of different commits. oops

PR #41, opened 2024-01-08
Mention possibility to donate in footer

Wituareard:


Wituareard:
Ok, I'm not sure about this anymore. It better describes the link, but when you see it on the front page without reading about the other possibilities, it can also look like that's the ultimate way to donate to the cause.

Wituareard:
Closing it for now

PR #42, opened 2024-01-09
Add Schema.org markup

Wituareard:
For reference see https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data and https://schema.org/BlogPosting

I also added the description meta tag, see https://developers.google.com/search/docs/appearance/snippet


joepio:
Great idea, thanks!

PR #43, opened 2024-01-09
Fix Schema.org markup

Wituareard:
The [Google Search Console ](https://search.google.com/test/rich-results/result/r%2Farticles?id=5JRts2_x2Sntp3qzWaBKnw)wants the image to have an `url` instead of a `contentUrl`, I don't know why I didn't catch that earlier.


joepio:
Thanks again! :) really appreciate all these PRs

PR #48, opened 2024-01-11
Improve defintion of intelligene in xrisk.md

Wituareard:
When talking about intelligence, people can mean a variety of different things. So I tried to make clearer that we don't want to share knowledge on the term itself, and instead just want to define what we are concerned about.


joepio (approved):

joepio:
Good!

Issue #45, opened 2024-01-10
Privacy policy


Wituareard:
As far as I know, the GDPR requires the site to have a privacy policy and some EU countries even require a cookie banner for using Google Analytics.

joepio:
You're right, we should add this.

Wituareard:
I'm not sure if only capturing users in the analytics who agreed to a cookie statement makes sense and also cookie banners are annoying, so it might be better to switch to something that doesn't require cookies.  The best free alternative I could find is [Cloudflare Analytics](https://www.cloudflare.com/web-analytics/).

Wituareard:
Okay, I've spoken to someone who knows this stuff a bit better than me, and analytics seem to require consent in general. You could use Cloudflare Analytics by routing the traffic through Cloudflare, which is fairly straight-forward to set up and also helps with performance. That'd probably still not be legal, but it's not as easy to get sued for (which would be a waste of time and money and was my motivation for bringing this up). That's because using Cloudflare in general doesn't require consent.

Wituareard:
[Audaxly.com ](https://audaxly.com/) seems like a good tool to generate a privacy policy, but it's something you have to do because it needs to contain a postal address etc

PR #66, opened 2024-01-15
Distribute fonts locally for performance and privacy

Wituareard:
Also helps with #45

Deploy preview: https://65a57d184e619b0008502ec5--frabjous-pasca-68e1cb.netlify.app/


joepio:
For me this makes performance a bit worse (Arc, MacOS)

Wituareard:
Hm okay, I didn't really test it, I've just read that it means -1 DNS request, SSL handshake etc. The main reason I came across this is that it makes the privacy policy easier.

joepio:
Makes sense. Let's go!

PR #71, opened 2024-01-16
Update remark-toc and add table of contents to FAQ

Wituareard:
Deploy preview: https://65a5cb8ae09f2500088f0b81--frabjous-pasca-68e1cb.netlify.app/faq


Wituareard:
Not the prettiest solution, but I figured that an FAQ where you can't jump to the question you have isn't really optimal, so it's probably better than nothing.

PR #73, opened 2024-01-17
Align external link icon to not make the line higher

Wituareard:
I don't know why "text-top" works to align it at the bottom, but it does across multiple browsers.

Before:
grafik

After:
grafik

Text:
grafik


joepio:
Thanks!

PR #91, opened 2024-02-09
Add Donate button on GitHub

Wituareard:
Needs to be enabled under Settings -> General -> Features -> Sponsorships

I don't really expect anyone to find the button here, but it takes 30 seconds, so why not.


joepio:
Great idea, thanks!

PR #95, opened 2024-02-14
Add full text search using Pagefind

Wituareard:
Two things I'm not entirely happy with:

- Search on a separate page instead of in a popup or something like that
- Needs to run `vite build` twice to include the files that are generated by Pagefind after the first build, but this only adds ~10 s which is about the same amount I optimized away by modifying the lucide-svelte import


joepio:
Great job! 

Regarding the separate page: is there a technical reason for this? Seems pretty doable to move it into a component to the navbar

Regarding the second `vite build` run: Is there a way we can change the config to build the search index in the first `vite build` run?

joepio:
Screenshot 2024-02-14 at 11 51 37

This looks absolutely amazing, great job!

Wituareard:
No technical reason, I just tried too long to build a decent looking and functional modal component and eventually gave up.

The problem is that the indexing would need to happen after the HTML files are generated, but before the netlify function is generated. Maybe that'd be possible with a vite or sveltekit plugin and a lot of investigation on how exactly things work, but it's probably not worth the effort.

The UI is mostly provided by Pagefind which is one reason I liked the library.

Wituareard:
Discord handle is dasSpieler

PR #97, opened 2024-02-19
Environmental draft Remmelt

joepio:
Written by Remmelt Ellen.

I'd love to hear some feedback about this page. It introduces a new risk and a new angle.

### DIFF
--- a/src/posts/environmental.md
+++ b/src/posts/environmental.md
@@ -6,94 +6,80 @@ description: AI's environmental impact is a problem not just because of the imme
 > ❝ The rise of the robots is a problem not just because of the immediate human cost, but
 > because robots are fantastically energy-hungry and thus accelerate the unsustainable,
 > insupportable damage we are doing to Earth.
- – Rupert Read, former spokesperson Extinction Rebellion
+> [Rupert Read](https://www.theguardian.com/theobserver/2016/mar/27/leters-robots-energy-consuming), past spokesperson of Extinction Rebellion
 
-Immediate harms are intense.
+Immediate harms are turning intense.
 
-Corporations feed our friends and family’s data into AI.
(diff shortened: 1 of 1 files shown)


Pato-desu:
Well, personally I like it for the most part but not the last sections, which I understand is what Remmelt would like to put the most.

When I wrote most sections of the /risks page, I try to look about the effects of AI on the climate and I didn't put it because I didn't find much and didn't feel comfortable written about something I didn't knew. Also considering that there's a lot of things that are hard to calculate, and narrow AI could bring pros on the environment which makes it net positive.

But I'm happy that Emmelt wrote about it and put sources. After reading the costs on the environment I feel more confident calling it a risk. 

But the last 2/3 sections read like the "disempowerment" and "x-risk" section again, and I think this page should exist mainly to convince climate activists. I feel like people who believe we could just turn-off the machines on the long run worry about not realizing it before too much damage has been done, not self-replicating machines.

P.D: I also kinda hate the video linked.

joepio:
> narrow AI could bring pros on the environment, AIs could end up being net positive for the environment.

I can see how AI might help in material science for a fusion reactor, for example, so I agree that there is a chance this all works out. But it depends on more assumptions than just "more AI = more compute used = more emissions", right?

> But the last 2/3 sections read like the "disempowerment" and "x-risk" section of /risks again, and I think this page should exist mainly to convince climate activists. I feel like people who believe we could just turn-off the machines on the long run worry about not realizing it before too much damage has been done, not self-replicating machines.

I think you're right. We'll probably lose some readers here who would otherwise be sympathetic.

joepio:
Removed the last part as per Remmelt's request

Pato-desu:
Which part did you removed @joepio ? Isn't the version live the original text?

PR #113, opened 2024-03-15
Outcomes page changes from Pato

Pato-desu:


Pato-desu:
So, besides not showing the probabilities, I wanted to redirect people to actions from their outcomes.

That led me to change the first questions to be specifically about a pause and add the idea we can change whether it happens or not.

One phrase that I added is **not serious at all**. I personally don't see the problem with it but let me know if you do. Also it could look a little repetitive if most answers end up in the same way, we could do a different sentence for each outcome.

Also I needed to remove the question about AI non SI being dangerous because it complicated things a lot, and giving how I changed the first question it didn't make sense.

Now I realized two problems with the preview:
- You cannot link stuff in the same way that in the articles.
- There's another button to show probabilities that I must remove too.

src/routes/outcomes/+page.svelte
```diff
@@ -18,9 +18,10 @@
 	{#if !intro}
 		<div class="headerbuttons">
 			<Button subtle on:click={() => (intro = true)}>Restart</Button>
+			<!--
 			<Button subtle on:click={() => (showProbabilities = !showProbabilities)}
-				>{showProbabilities ? 'Hide' : 'Show'} probabilities</Button
-			>
+				>{showProbabilities ? 'Hide' : 'Show'} probabilities</Button>
```
joepio, line 23: Hiding the button is probably fine as a temporary solution.

src/routes/outcomes/tree.ts
```diff
@@ -16,6 +16,9 @@ type outcome = {
 	text: string
 }
 
+const doomer_hope_str = " Luckily, you've just learn this, as if it were a message from the future! And you are still on time to make a Pause happen. Good luck John Connor, the future is not set."
```
joepio, line 19: I think the messages should link to an action page. Instead of adding it to the tree I'd add it to the view I think

src/routes/outcomes/tree.ts
```diff
@@ -44,18 +47,18 @@ const humanTakeOver: TreeNodeType = {
...
-			text: 'Even though extremely powerful AI systems exist, AIs never go rogue, nobody succeeds in using AI to take over control and AI never causes large-scale catastrophes. We hope this is the universe we live in, but it seems like wishful thinking.'
+			text: 'Even though extremely powerful AI systems exist, AIs never go rogue, nobody succeeds in using AI to take over control and AI never causes large-scale catastrophes. We hope this is the universe we live in, but it seems like wishful thinking.' + optimist_help_str
 		}
 	}
 }
 
 export const tree: TreeNodeType = {
 	type: 'question',
-	text: "Will we continue to build AI until it's smarter than humans?",
+	text: "Will we continue to build AI without pausing?",
```
joepio, line 61: nice

src/routes/outcomes/tree.ts
```diff
@@ -85,35 +88,24 @@ export const tree: TreeNodeType = {
...
-		yes: {
-			type: 'doom',
-			text: 'An AI-related catastrophe happens before we even get to superintelligence. Since some dangerous capabilities do not even require general superintelligence, this might be one of the more likely outcomes.'
-		},
-		no: {
-			type: 'safe',
-			text: "If we successfully prevent dangerous AI from being developed, it can never outsmart humans, and we can even prevent other forms of AI catastrophe. We hope we're living in this universe. You can help to achieve this!"
-		}
+		type: 'safe',
+		text: "If we successfully prevent dangerous AI from being developed we can prevent AI catastrophes. We hope we're living in this universe. You can help achieve this!"
```
joepio, line 109: Yeah this is probably more consistent with our proposal / ask. But it may feel a bit disingenuous - I do think we're still at risk of disaster even if we pause!

src/routes/outcomes/TreeNode.svelte
```diff
@@ -44,8 +44,8 @@
 	{#if showInfo}
 		<div>
 			<p>
-				{node.explanation} PauseAI estimates this probability at roughly
-				<b>{node.probability * 100}%</b>.
+				{node.explanation} <!--PauseAI estimates this probability at roughly
```
joepio, line 47: I think we could add an if condition that checks the `type` of the node, and conditionally prints the `doomer_help_str` or `optimist_help_str`. It would be cleaner than adding the text to the source json.

joepio (changes requested):

src/routes/outcomes/TreeNode.svelte
Pato-desu: I prefer the particular message that the safe node in case of a pause has. Ideally they should all be personalized.

src/routes/outcomes/tree.ts
Pato-desu: sure, it was like that. I changed the can for could now.

joepio:
Hmm I think we should reverse the changes, because this indeed is not better than our first version. Just got this feedback:

> The AI Outcomes app (https://pauseai.info/outcomes) found under the P(doom) page reduces credibility of AIPause.info in my opinion by immediately asking the user to input binary yes/no answers to questions that a person obviously should have uncertain answers to. Clearly this mean that the app can't lead to a good estimate of one's p(doom) and isn't a good procedure for estimating p(doom).

Pato-desu:
I mean I agree with the edit in your message. It was never a pdoom calculator.

PR #127, opened 2024-03-25
Update Jaan Tallinn quote

mrbreastly:
Add attribution info for the included image... " "© Annika Haas, https://estonia.ee/"


Pato-desu:
Does it say it's fair use if we just put the name? Just because it says their name in the photo I think you can't assume that. I think we must use this website https://openverse.org/search/image?q=jaan%20tallinn&license_type=commercial,modification

PR #128, opened 2024-03-25
Update Paul Christiano quote

mrbreastly:
- Add attribution for the image... "© Paul Christiano, https://paulfchristiano.com"


Pato-desu:
Same as with the Tallinn PR. I don't know if it's enough with just credit them. Look at his Wikipedia page, he doesn't have any picture.

Issue #164, opened 2024-05-06
Prettier formatting?


moiri-gamboni:
Hi, 
I noticed you have prettier set up but many of the files were not formatted. Is there a reason? 
I can submit a PR for a pre-commit hook if you want.

joepio:
No particular reason. Sounds great!

PR #165, opened 2024-05-07
Add a pre-commit hook to format files with Prettier

moiri-gamboni:
This PR:
1) Formats existing files with Prettier.
2) Installs husky and lint-staged.
3) Adds a pre-commit hook so that only staged files are automatically formatted with Prettier.

Additionally, I cleaned up the `.prettierignore` file, as prettier automatically reads `.gitignore`, and added an ignore entry for a twitter widget that was copied from their website. I also removed the `--plugin-search-dir` flag in `package.json` as it isn't used in prettier v3

closes: #164


joepio:
LGTM, thank you for this!

PR #172, opened 2024-05-26
Format markdown

tylerlaprade:
* `bun run format` to autoformat all files
* Added extensions for shared tools to the workspace recommendations


RisingOrange:
Closing as stale. We've moved from Bun to pnpm (so no more `bun.lockb`), and `.vscode/extensions.json` is already updated in main. Thanks though!

tylerlaprade:
@RisingOrange No worries, thanks for the update! Just curious for my own education what are the benefits of PNPM over bun?

RisingOrange:
@tylerlaprade Good question, I'm not sure what the reasoning was for choosing pnpm, I wasn't involved in that.

anthonybailey:
@joepio or @Wituareard welcome to confirm how bun was used, but we definitely don't need two package managers. Current state looks fine. Claude's guess from git history sounds right to me.

"""
Bun was never project tooling. No README, script, or CI ever referenced it. bun.lockb entered the repo on 2023-12-15 inside Joep's "Local communities" content commit — classic lockfile leakage from a contributor using bun locally. Same thing happened again with Deno: deno.lock arrived 2025-02-07 inside "Add photoresist", also Joep's.

The removal was deliberate and reasonable. Commit 6f45b10 (Wituareard, 2025-03-10, "Remove lockfiles #287") deleted both bun.lockb and deno.lock and added them to .gitignore under a comment "alternative package managers". 

Current state is textbook-correct:
pnpm-lock.yaml has been there since the repo's first commit (2023-04-28) — pnpm was always the canonical manager.

packageManager: pnpm@11.1.3 is pinned in package.json, so corepack enforces it.
CI (pnpm-check.yml) runs pnpm install --frozen-lockfile, which fails if the lockfile drifts.

The gitignore approach is a good compromise: contributors may use bun/deno locally, but their lockfiles can't pollute the repo.

Wituareard:
I don't think pnpm was ever actively chosen since the repo originates from a template. Explanation regarding bun looks about right

PR #174, opened 2024-05-29
Fix typo in growth-strategy

c-riq:
Really nice resource.
I think it should be `contribute` without the `s` as in `should contribute`, right?

Aside from the typo, I thought to myself `Be a teacher` could have a bit of a negative connotation as people may not want unsolicited teaching.
An alternative wording could be `Help others grow`. What do you think?


Pato-desu:
good suggestions, sorry to take 3 weeks to seeing this lol

PR #186, opened 2024-07-20
Fix markdown imports

moiri-gamboni:
Remove `export {}` from `app.d.ts` to keep type declarations in the global scope. This allows `.md` file type declarations to be recognized project-wide, fixing module resolution errors for Markdown imports.

Otherwise, doing:
`import MyPost from '../../posts/mypost.md'` throws `Cannot find module '../../posts/mypost.md' or its corresponding type declarations.ts (2307)`


joepio:
Thank you!

PR #196, opened 2024-08-21
August Post Updates

hydraseek-pauseai:
Changed wording of title so it would appear better in URL previews. Also added Twitter since it's not empty.


joepio:
I added the changes, thanks!

PR #202, opened 2024-09-15
Modify lucide import to speed up build

Reginred:


joepio:
Hi! Thanks for this PR. Can you help me understand why this speeds up the build?

Reginred:
With the current import the bundler has to process one big file with all icons and the new import references a file which only contains the individual icon.

joepio:
Thank you!

PR #207, opened 2024-09-18
Bump lockfile following pnpm update

anthonybailey:
Having set up locally, my pnpm-lock.yaml shows a large diff,
 starting with lockfileVersion bumping from 6.0 to 9.0.
But Joep says
> Looks like pnpm has recently been updated
> You can commit the lockfile change!

Well, OK! My first commit here, so let's hope it isn't a disaster.


joepio:
Great, thanks!

PR #210, opened 2024-09-20
Exclude prerendered routes from edge function

Reginred:
I patched the Netlify adapter to make the edge function only handle dynamic routes, which seems to pretty consistently improve performance (including some metrics that currently "need improvement" according to PageSpeed Insights), partly because it allows Netlify to handle caching. It also reduces the number of billed function calls (which is probably negligible right now, but still) and generally seems like a better way to do things.

It would be good to test the performance on the paid account and generally make sure everything works via a Deploy Preview before merging. I'll put links to the deployments I used for testing below.

Discussion on Discord: https://discordapp.com/channels/1100491867675709580/1226136907046588476/1281710876549709836

Preview on free account:
https://adapter-netlify-selective--deft-starlight-2e4321.netlify.app/

PageSpeed Insights for preview on free account:
https://developers.google.com/speed/pagespeed/insights/?url=https://adapter-netlify-selective--deft-starlight-2e4321.netlify.app/

Current approach on free account:
https://deft-starlight-2e4321.netlify.app/

PageSpeed Insights for current approach on free account:
https://developers.google.com/speed/pagespeed/insights/?url=https://deft-starlight-2e4321.netlify.app/

### DIFF
--- a/src/lib/adapter-netlify-selective.js
+++ b/src/lib/adapter-netlify-selective.js
@@ -1,0 +1,31 @@
+import adapter from '@sveltejs/adapter-netlify'
+import fs from 'fs/promises'
+
+const NAME = 'adapter-netlify-selective'
+const MANIFEST_PATH = '.netlify/edge-functions/manifest.json'
+const FUNCTION_NAME = 'render'
+
+export default function ({ split = undefined, edge = undefined } = {}) {
+	const adapterNetlify = adapter({ split, edge })
+	adapterNetlify.name = NAME
+	const adaptNetlify = adapterNetlify.adapt
+	adapterNetlify.adapt = async function (builder) {
+		await adaptNetlify(builder)
+
+		if (!edge) throw new Error(`${NAME} only works for edge functions`)
+
+		const dynamicRoutes = builder.routes.filter((route) => !route.prerender)
+		const dynamicRoutesPattern = dynamicRoutes
+			.map((route) => '(' + route.pattern.source + ')')
+			.join('|')
+
+		/** @type import("./types").EdgeManifest */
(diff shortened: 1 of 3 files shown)


anthonybailey:
I get the intent but have no idea about Netlify details.

Is it possible to show (and where non-obvious, explain how) how your new adapter diffs from the (presumed default) @sveltejs/adapter-netlify ?

Reginred:
It calls the original `build` function of `adapter-netlify` before doing anything, so nothing is removed, but it overrides the default pattern `^/.*$` (matches all requests) with one that is composed of the patterns for the dynamic routes. This shouldn't impact performance too much if Netlify properly compiles simple regexs like this, which the tests seem to support.

Reginred:
Alright, I just noticed that this breaks the 404 page. Back to the drawing board.

anthonybailey:
Right, I see: this updates the .netlify/edge-functions/manifest.json file
on disk. I assume the build runs on start up or deploy or similar (netlify
noob here - any overview in standard docs re where builds feature in
deploy/start up and what happens on the network when pages are served could
be handy.)
Can you copy in here the new output that gets logged by the builder?

How does that new EdgeManifest type get used?

You expect this to avoid a bunch of calls to the server, right? The two
PageInsights reports look near identical so I assume they don't know about
any Netlify caching upstream(?) How have you confirmed it works as intended
/ measured the time/calls saving?

Final dumb question: what's the approx proportion of dynamic to cacheable
pages? (Ideally, both number of different routes, and scaled by page
popularity i.e. across typical traffic.)

Push back if any of this is stupid or difficult: as advertised, I'm coming
at this fresh. I'm asking mostly because it will interplay with pending
localization plans.

Thanks!

Issue #157, opened 2024-04-21
Offer translations


Wituareard:
Discord discussion: https://discord.com/channels/1100491867675709580/1226136907046588476

joepio:
This is a very big and important question, as there exists little AI risk knowledge in most non-english languages. We need to get this right. 

## Wants

- Translate to a lot of languages without too much effort (would be great if we can utilize AI to translate)
- Keep pages up to date. We need a process for this, expect automated translations to require QC / fine-tuning.
- Have custom pages for certain nationalities that are _not_ translated.

## Technical approaches

### [sveltekit-i18n](https://github.com/sveltekit-i18n/lib)

joepio:
@imotaru

Imotaru:
- Having a button to change between different languages on the website (good example is how Finnish government sites does it e.g. https://valtioneuvosto.fi/en/decisions).
- Keep track of if a page was manually translated / verified or not. If the translation has not been manually verified could have a note "this page is a machine translation, if you want to help improve the translation click here (could just be a link to discord or to email me or someone else on the team if we want to keep it simple)
- Version number could indicate how important the change is, like 1.0.1 to 1.0.2 is low prio, but 1.0.1 to 2.0.0 is a big change and high prio. Low prio change would be something like a typo fix that only affects English anyways, and high prio would be something that adds very important info or if the old info does not apply anymore / will be misleading.

joepio:
@Imotaru Great insights.

- The switch button should probably be a dropdown
- Some versioning system sounds like a good idea. Maybe we can use something like `v1.2ai` to indicate if this has been AI generated, whereas `v1.2m` is a manually checked one. I think we could put this in the md metadata

joepio:
## Current idea

- Have a JS script that iterates over every markdown page and creates translations for every configured language using an LLM
- Use Git to do version management
- Every language can have an additional set of instructions to inject in the prompt
- Every translated page can have an addiontal page-language specific prompt
- Store if files are generated in the document

anthonybailey:
OK. So Joep's plan and the one I'd arrived at independently are pretty similar and even complementary.

My evolving ideas are captured in [the Discord Project Babel thread](https://discord.com/channels/1100491867675709580/1226136907046588476/1226136907046588476) and at the time of writing conclude with a "Sonnet 3.5, tell me why I'm stupid about this" interaction linked from [here](https://discord.com/channels/1100491867675709580/1226136907046588476/1281201307256094791).

[...]

PR #211, opened 2024-09-22
Offer translations #157 - Script + NL

joepio:
Toyed around a bit with translations #157. Not sure if this approach is the best.

Also, help is very much welcome!

- [x] Script that generate translations, iterates over all posts
- [x] Basic language configuration
- [x] Language routing
- [x] Language switcher
- [ ] Include other languages
  - [x] NL
- [ ] Translated non-post pages
  - [x] Home
- [ ] Translated components (footer, navbar)
- [ ] Versioning in translated files 
- [ ] AI generated notice


anthonybailey:
Joep, I'm refining detail of a prompt myself. I've run a fair few
experiments. We should exchange notes.

I really want to pair on this stuff because in passing I'll learn a lot
about the design history of the website.

What do you think about my breakdown above: do you agree that most page
text content that is not in markdown (e.g. in .svelte, occasional fragments
in .ts, and some structured pieces) can be?

joepio:
> do you agree that most page
text content that is not in markdown (e.g. in .svelte, occasional fragments
in .ts, and some structured pieces) can be?

@anthonybailey I don't think that's a good solution, at least for most of the items. I think we should have a bunch of json files.

anthonybailey:
Right. So you want to plug the "standard" sveltekit i18n but power it from LLM. I get that. Do you have a proposal re exactly how routing will work?

I don't want to get in your way, but I have gone into the weeds re how to make the ongoing auto-LLM work, and I really do want to learn about the website from you.

The code Claude is accelerating may work but it is spaghetti and going to make maintenance hard. I can embrace the future but we should be factoring this stuff. Our new apprentice has a huge amount of energy but in September 2024 they need more design oversight.
Please can we arrange some time to work together on this?

Reginred:
A fundamental problem with this approach is that pages using the search params [can't be prerendered](https://kit.svelte.dev/docs/page-options#prerender).

joepio:
> @joepio Is there a specific reason why you added manual splitting of the path instead of using SvelteKits folder-based routing? The language could also be made an optional parameter to continue serving the English content under the current paths (and maybe redirecting from /en).

I think I used folder based routing in the last few commits. As in, use a `{lang}/{slug}` directory structure.

Maybe I'm bit understanding what you're saying / missing something!

joepio:
I'm not planning on working on this PR anymore, this new one will be the one to watch: https://github.com/PauseAI/pauseai-website/pull/226

PR #212, opened 2024-09-24
Events page luma

joepio:
- [x] Add `/events` page 
- [x] Replace `/protests` in the navbar with `/events`
- [ ] Show event lists in communities

### DIFF
--- a/src/posts/events.md
+++ b/src/posts/events.md
@@ -1,0 +1,21 @@
+---
+title: PauseAI Events
+description: Meetings, workshops, protests, and more!
+---
+
+<iframe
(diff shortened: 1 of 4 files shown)


joepio:
@anthonybailey

anthonybailey:
Change looks sane in design and implementation, would accept if it wasn't for the error.

The build logs are pretty noisy but looks like this is just still linking the /protesting route from roadmap.

I would suggest 301 redirecting /protesting to /events as well as any internal fixes we make.

@joepio not clear whether debate practice was just a handy example or a special case - would we end up putting all open Discord meetings here?

If not, I guess a link to the calendar within Discord would be sensible.

PR #217, opened 2024-10-03
Shortening and polishing the /action page

Pato-desu:


Pato-desu:
if there's problems is maybe because of how I located the comments

src/posts/action.md
```diff
@@ -4,61 +4,32 @@ description: Ways to help out with pausing AGI development.
...
 
 ## Join the movement
 
-- **[Join PauseAI](/join)** and help us grow.
-- **Join our [Discord server](https://discord.gg/T3YrWUJsJ5)**, where our community is most active. We have a **#projects** channel where people are working on campaigns, videos, images, apps and more.
-- **Join one of our [teams](/teams)** and collaborate with other motivated volunteers.
-- **Meet** other people in a [PauseAI community near you](/communities). If there is no community near you, [starting one](/local-organizing).
+- Join or create a [local community](/communities) or [event](/events). 
+- Join the PauseAI [Discord](https://discord.gg/2XXWXvErfA), where most of the collaboration happens.
+- If you want to volunteer, fill out [this form](https://airtable.com/embed/appWPTGqZmUcs3NWu/pagoxRuCai4OYJEHt/form).
```
joepio, line 28: This is a radical change, since pretty much the entire website converges to `/join`

anthonybailey: How do a radical change to said key page? It just adds the `/events` link though, doesn't it?

Pato-desu: > How so a radical change to said key page? It just adds the `/events` link, doesn't it?

Nono, that's the change to the /join page. The change Joep pointed up to is in the /action page.

It's better to reduce the amount of clicks, no? If not why don't we just put "[**Join us**](/join)"?
The present section is really repetitive and confusing.

anthonybailey:
Ah, so replacing the link to join with separate events and community links
may break some conventions that e.g. inform how we measure conversion?

I agree the distinct "Join" and "Act" pages with somewhat overlapping
items  seems a bit tangled. Not clear how to collapse the pages though.

PR #218, opened 2024-10-05
Load Google Analytics in web worker using Partytown

Wituareard:
The library [Partytown ](https://partytown.builder.io/) enables loading Google Analytics in a web worker, which can significantly improve website performance and as a result boost search engine rankings.

According to their website, Partytown is still in beta because it doesn't yet support arbitrary scripts. But for Google Analytics and other popular analytics tools it's already being used by companies like Notion, Upwork, Walmart and ExpressVPN ([list](https://www.wappalyzer.com/technologies/performance/partytown/)).


anthonybailey (approved):
Looks great to me. I suggest we don't use Partytown for anything else yet, but agree that wrapping Google analytics in particular seems well-established, and the net change to performance measurements is compelling.

PR #219, opened 2024-10-05
Cache people page

Wituareard:


anthonybailey:
It's great that folk are contributing!

I'm in a bit of a bind here. I'm wanting to take load off Joep (who made the website design choices and is familiar with relevant frameworks and tradeoffs) and to act usefully as software lead (reviewing, looking for problems, nudging us in sensible directions given our expected value in the future.) I'm coming in with experience, but not of this particular stack.

It seems likely that anyone contributing has good ideas and understanding of same choices and tradeoffs, and their efforts are hugely appreciated.

Even so: I lack the context needed to do useful review. I grok the two commit headers. I see the shape of the change. But I can't assess whether it is sensible or has longer term cost without a bit more help and explanation.

Top-level: how did these pieces play together before the change? When a client with/without browser cache requested the /people URL, what network requests flowed through Netlify / Svelte / Airtable? Why do we want to make the change? I assume we want to reduce Airtable traffic and cost, and believe it will also interplay with localization routing ideas.

Part of the curiosity is that rather a lot of files were touched. Naively I would have expected this kind of choice to have been expressible more simply within Svelte.

Knowing what testing choices were made would also make me more confident!

Sorry for starting from ignorance here!

anthonybailey:
Thanks - especially for the docs link. Ok, so this is "edge" in the CDN sense i.e. our Svelte website gets no traffic so long as Netlify itself is happy. And because Netlify manages our deploys, it is the case that we can clear bad caches by deploying. (So e.g. if we fixed something in Airtable that was important to rush out before the 1h TTL, a bounce deploy lets that happen.)

Joep liked that the people page order was randomized (to underline how every volunteer is important) but changing the order once per hour seems sufficient. And sure, caching the page in the edge CDN certainly lightens the load on Airtable without introducing any cache in the Svelte website itself. Great.

(Shadowing concern noted.)

I'll also dig into https://docs.netlify.com/frameworks/sveltekit to catch up some on all this.

joepio:
Thanks for this!

PR #222, opened 2024-10-09
Simplify homepage styling for translation

Wituareard:
For a lot of solutions to translation, including HTML tags in messages seems to be a lot easier than including components. But I wonder if there is a reason to strongly prefer the current solution, because it is also a lot more complex internally, and maybe there is a reason for that.


joepio:
Good idea, thanks!

PR #224, opened 2024-10-21
Some proposed changes to the Summit page

Pato-desu:
* divided in coming and past summits
* moved coming to the top
* polished (IMO) text about the SF summit


joepio:
Thanks pato!

PR #225, opened 2024-10-23
Backfire Correction

UFO-101:
Correct inaccurate claim that protests have no "backfire" effects

My friend who previously worked at Social Change Lab reached out to note that there is substantial evidence for backfire effects from protests, that the linked study even acknowledges.

### DIFF
--- a/src/posts/faq.md
+++ b/src/posts/faq.md
@@ -76,7 +76,7 @@ Check out our [theory of change](/theory-of-change) for a detailed overview of o
 - Protesting shows the world that we care about this issue. By protesting, we show that we are willing to spend our time and energy to get people to listen.
 - Protests can and often will [positively influence](https://www.socialchangelab.org/_files/ugd/503ba4_052959e2ee8d4924934b7efe3916981e.pdf) public opinion, voting behavior, corporate behavior and policy.
 - By far [most people are supportive](https://today.yougov.com/politics/articles/31718-do-protesters-want-help-or-hurt-america) of peaceful/non-violent protests
-- There is [no "backfire" effect](https://journals.sagepub.com/doi/full/10.1177/2378023120925949) [unless the protest is violent](https://news.stanford.edu/2018/10/12/how-violent-protest-can-backfire/). Our protests are peaceful and non-violent.
+- There is evidence that "backfire" effects from protesting [are limited](https://journals.sagepub.com/doi/full/10.1177/2378023120925949) [unless the protest is violent](https://news.stanford.edu/2018/10/12/how-violent-protest-can-backfire/). Our protests are peaceful and non-violent.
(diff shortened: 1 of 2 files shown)


UFO-101:
I wasn't referring to the new results of this paper, but other papers which this study cites

> For example, studies have shown that providing conservatives and Republicans with climate information not only may have no effect but can instead produce “backfire” effects wherein support for climate policy or belief in anthropogenic climate change is actually reduced because of exposure to conflicting information (e.g., protests) ([Dixon, Bullock, and Adams 2019](https://journals.sagepub.com/doi/full/10.1177/2378023120925949#bibr19-2378023120925949); [Hart and Nisbet 2012](https://journals.sagepub.com/doi/full/10.1177/2378023120925949#bibr40-2378023120925949); [Zhou 2016](https://journals.sagepub.com/doi/full/10.1177/2378023120925949#bibr83-2378023120925949)).

> Given the partisan nature of climate change belief and the evidence for backfire effects in climate communication

joepio:
The study you cite refers to providing climate information leading to backfire effects. Does this also apply to protests? Are there other studies measuring a backfire effect from protests?

Seems good to keep it factual in any case.

joepio:
From. The studies discussion section:

> Despite the plausible concern that certain forms of protest may cause the type of backfire effect observed in climate communication research, I find no evidence to justify that concern.

I feel like saying there are limited backfire effects is not an accurate reflection if there is no evidence.

Perhaps we should state "there is no evidence of backfire effects from protests"

UFO-101:
From my friend who researches in this area

> Will message in morning. Generally I feel pretty certain that it is well known that protests certainly CAN backfire. There’s a case study on the SCL website I did on the group SHAC, where it seemed very apparent that protest backfired. And SCL also did an RCT where most protest descriptions led to people disliking the protest group: https://www.socialchangelab.org/_files/ugd/2ed91c_35d9cc4567c849cbba9bab51ee5efa25.pdf

> Now technically those are all mostly more disruptive than any previous Pause stuff (AFAIK), but I wouldn’t classify them as violent.

Let's wait till tomorrow, but I think it might be fair to update it to say:
```
There is little evidence for "backfire" effects from protests, unless the protests are highly disruptive or violent
```

joepio:
I think saying little evidence suggests that there is some. 

How about "there is no evidence of backfire from peaceful, non-disruptive protests."

UFO-101:
I think little evidence is good. I don't have time to do a literature review, but I'm sure there are studies showing a backfire effect.

joepio:
I changed the line to 

`There is no evidence for a "backfire" effect unless the protest is violent` which accurately reflects the literature that I've found. If anyone can point me to evidence pointing in a different direction, I'll update.

PR #240, opened 2024-11-28
New Frontpage quote + smaller pictures for quotes

Pato-desu:


Pato-desu:
> The high resolution backgrounds are useful for the downloads. Ideally, we'd use enhanced:img to generate multiple resolutions and formats and let the browser choose a different one for regular display and high-resoltion download. But I'm not entirely sure how to implement that with the background-image property. And also, the lower resolutions alone don't seem to reduce the file size that much.

The resolutions of the downloads are always 2000x1000, so there is some resolutions that are too big even for downloads. Also, pictures are probably not even being downloaded.
The sizes of some of them were reduced by 90%+.

> Also, I think "Nobel Prize winner says it might be sensible to stop developing further" supports our proposal better than "Anonymous experts say xrisk is a global priority". But for decisions like that, it would be useful if the comms team could weigh in which @anthonybailey wanted to take care of.

Maybe. In the picture I put the 4 most important labs, which I expect to be the relevant part. I think we should cut then some experts which are not known by the public (e.g.: Stuart Russell) and maybe add politicians.

We could also quote the FLI letter and clarify below that we think it should last more than 6 months.

joepio:
Screenshot 2024-11-29 at 12 17 45

Hmm we should have a better image here. Not sure if this works in the context of quotes IMO.

joepio:
The `@glidejs/glide/dist/glide.modular.esm` import fails.

```tsimport Glide, { Controls, Images, Keyboard, Swipe } from '@glidejs/glide/dist/glide.modular.esm'
```

```
12:20:35 PM: Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/opt/build/repo/node_modules/@glidejs/glide/dist/glide.modular.esm' imported from /opt/build/repo/.svelte-kit/output/server/entries/pages/_page.svelte.js
12:20:35 PM:     at new NodeError (node:internal/errors:405:5)
12:20:35 PM:     at finalizeResolution (node:internal/modules/esm/resolve:327:11)
12:20:35 PM:     at moduleResolve (node:internal/modules/esm/resolve:980:10)
12:20:35 PM:     at defaultResolve (node:internal/modules/esm/resolve:1206:11)
12:20:35 PM:     at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:404:12)
12:20:35 PM:     at ModuleLoader.resolve (node:internal/modules/esm/loader:373:25)
12:20:35 PM:     at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:250:38)
12:20:35 PM:     at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:39)
12:20:35 PM:     at link (node:internal/modules/esm/module_job:75:36) {
12:20:35 PM:   url: 'file:///opt/build/repo/node_modules/@glidejs/glide/dist/glide.modular.esm',
12:20:35 PM:   code: 'ERR_MODULE_NOT_FOUND'
12:20:35 PM: }
```

Maybe it will work with the netlify adatper.

Pato-desu:
> @Pato-desu
> 
> > The sizes of some of them were reduced by 90%+.
> 
> I don't see significant reductions for files that aren't a single color under "Files changed"

At the contrary, those are the least reduced, if I understand correctly ?
But I'm also assuming some stuff about how the memory allocation / loading of the page works. I created the high res folder thinking that those pictures don't really get loaded, so I'm ignoring them in my analysis.

Pato-desu:
> mm we should have a better image here. Not sure if this works in the context of quotes IMO.

I could either reduce the picture or put Sam Altman

Pato-desu:
I have asked the Comms team about their opinion on the preview and the quotes in the frontpage. If they like it I'll merge it

Issue #57, opened 2024-01-12
Add to SOTA: worse than humans at agency


Pato-desu:
To be a little more honest, and sympathize with skeptics. We can add there that even if LLMs right now aren't agentic the big labs are trying to change that.

joepio:
Yeah the SOTA needs some more things AIs aren't good at. However, I don't think agency is that clear to me. The reason that AutoGPT doesn't really lead to useful behaviour has to do with a bunch of underlying shortcomings:

- Short context window (little short-term memory)
- Hallucinations
- Can't update own weights (learn as you go).

Pato-desu:
I don't know if I agree with you, but I guess agency can mean two different things. One is acting like an agent and the other deciding stuff somewhat independently. Right now they are "bad" for both.

joepio:
What does it mean to "act like an agent", though? I just don't have a clear understanding of this concept I suppose.

We have NN's playing all sorts of video games. I'd consider these agents. They perform actions in a system to achieve some objective. 

I agree that if you give AutoGPT a goal, it performs sub-human in pretty much any task. But I think the reason it's bad can be explained mostly by other shortcomings: it can't make a cup of coffee because it can't control a body. It can't reliably perform tasks as it hallucinates things. 

But maybe I'm missing some important other dimension here?

Pato-desu:
Yeah, you're right. I'm not sure what I wanted to refer to. I'm confused now.

Pato-desu:
Being general purpose?

joepio:
Still not clearly defined to me.

Pato-desu:
> I agree that if you give AutoGPT a goal, it performs sub-human in pretty much any task. But I think the reason it's bad can be explained mostly by other shortcomings: it can't make a cup of coffee because it can't control a body. It can't reliably perform tasks as it hallucinates things.
> 
> But maybe I'm missing some important other dimension here?

Why is the reason why is bad relevant? 
Do you think "Accomplishing instructed goals" is a good term for the thing we are referring to?
Also, they are bad at "Coming up with goals" IMO. So I would prefer if we had something like "Accomplishing goals for which they were not trained"

PR #244, opened 2024-12-07
some proposed changes to the state of the art page

Pato-desu:
## Explanation behind the changes
* I don't think that AIs "think":
It just doesn't fit. I'm not sure if it feels right even with AGI-level robots on stories. And I think other people feel this way. I do not consider that animals think, and the same applies to AIs. 
* Storage efficiency is not a capability... idk it just doesn't fit and who cares? If the models had more weights than humans have synapses in their brains it wouldn't matter.
* Language sounded weird so I divided it into "Writing" and "Translating".
* I could consider them being "smarter"/ or "having IQ", but they don't seem to have something like system 2 thinking, although AGI companies are working to change that. When they "explain their reasoning" a lot of the time they don't make absolutely any sense. They also do worse than humans at https://simple-bench.com/.
* I added the goals point that I mentioned here https://github.com/PauseAI/pauseai-website/issues/57

##
I can change my mind about them, but the question is, do we think we win/convince more people than the ones we lose by using controversial terms?


src/posts/sota.md
```diff
@@ -39,6 +38,7 @@ _Last update: 2024-09-16_
...
 - **Continual learning**. Current SOTA LLMs separate learning ('training') from doing ('inference'). Although LLMs can learn using their _context_, they cannot update their weights while being used. Humans learn and do at the same time. However, there are multiple [potential approaches towards this](https://arxiv.org/abs/2302.00487). A [2024 study](https://arxiv.org/html/2402.01364v2) detailed some recent approaches for continual learning in LLMs.
+- **Accomplishing goals** for which they were not trained and are not about outputting text, images or audio. If you ask an LLM with control over a computer or a robot, to do not-super-easy actions, or accomplish goals in virtual or real environments far from its training distribution, it will likely fail.
```
joepio, line 41: Good addition

src/posts/sota.md
```diff
@@ -14,17 +14,16 @@ _Last update: 2024-09-16_
...
-- **Reasoning**: o1 [correctly answers 78%](https://openai.com/index/learning-to-reason-with-llms/) of GPQA diamond questions, outperforming human domain experts (PhDs) who only get 69.7%.
+- **Writing**: In December 2023, an AI-written novel won an award at a [science fiction national competition](https://www.scmp.com/news/china/science/article/3245725/chinese-professor-used-ai-write-science-fiction-novel-then-it-won-national-award?campaign=3245725&module=perpetual_scroll_0&pgtype=article). The professor who used the AI crafted the narrative from a draft of 43,000 characters generated in just three hours with 66 prompts. The best language models have superhuman vocabulary and can write in many different styles.
+- **Translating**: And they can respond and translate to all major languages fluently.
```
joepio, line 23: Removing the "reasoning" section is dubious IMO. It's an important skill that we need to list here.

src/posts/sota.md
```diff
@@ -14,17 +14,16 @@ _Last update: 2024-09-16_
 - **Games**: For many games ([Chess, Go](https://en.wikipedia.org/wiki/AlphaGo_Zero), Starcraft, Dota, [Gran Turismo](https://www.technologyreview.com/2022/07/19/1056176/sonys-racing-ai-destroyed-its-human-competitors-by-being-nice-and-fast/) etc.) the best AI is better than the best human.
 - **Working memory**: An average human can remember about 7 items (such as numbers) at a time. Gemini 1.5 Pro [can read and remember 99% of 7 million words](https://blog.google/technology/ai/google-gemini-next-generation-model-february-2024/#sundar-note).
 - **Reading speed**: A model like Gemini 1.5 Pro can read an entire book in 30 seconds. It can learn an entirely new language and translate texts in half a minute.
-- **Thinking speed**: AI models can write at speeds far surpassing any human, writing entire computer programmes in seconds.
+- **Writing speed**: AI models can write at speeds far surpassing any human, writing entire computer programmes in seconds.
```
joepio, line 17: This seems unfair / seems to undersell their advantage. Yes, humans are slow at writing because of pens and keyboards and mechanics - but that's not why these LLMs speed is impressive. It's because they can output thoughts / language / text at a far greater speed than we can

src/posts/sota.md
```diff
@@ -14,17 +14,16 @@ _Last update: 2024-09-16_
...
 - **Amount of knowledge**: Modern LLMs know far more than any human, its knowledge spanning virtually every domain. There is no human whose knowledge breadth comes close.
-- **Storage efficiency**: GPT-4 has about [1.7 trillion parameters](https://the-decoder.com/gpt-4-architecture-datasets-costs-and-more-leaked/) (neuron connections), whereas humans have [about 100 to 1000 times the number of synapses](https://www.jax.org/news-and-insights/jax-blog/2018/December/600-trillion-synapses-and-alzheimers-disease) (neuron connections). However, GPT-4 knows thousands of times more, storing more information in a smaller amount of parameters.
-
+
```
joepio, line 19: I think this is an important insight, why remove it?

joepio: hmm to be fair you're right that it's not really a capability.

src/posts/sota.md
```diff
@@ -14,17 +14,16 @@ _Last update: 2024-09-16_
...
-- **IQ**: With verbal IQ tests, LLMs have been outperforming 95 to 99% of humans for a while (score between [125](https://medium.com/@soltrinox/the-i-q-of-gpt4-is-124-approx-2a29b7e5821e) and [155](https://www.scientificamerican.com/article/i-gave-chatgpt-an-iq-test-heres-what-i-discovered/)). With non-verbal (pattern matching) IQ tests, the 2024 o1-preview model scored [120 on the mensa test](https://www.maximumtruth.org/p/massive-breakthrough-in-ai-intelligence), beating 91% of humans.
+- **IQ tests**: With verbal IQ tests, LLMs have been outperforming 95 to 99% of humans for a while (score between [125](https://medium.com/@soltrinox/the-i-q-of-gpt4-is-124-approx-2a29b7e5821e) and [155](https://www.scientificamerican.com/article/i-gave-chatgpt-an-iq-test-heres-what-i-discovered/)). With non-verbal (pattern matching) IQ tests, the 2024 o1-preview model scored [120 on the mensa test](https://www.maximumtruth.org/p/massive-breakthrough-in-ai-intelligence), beating 91% of humans.
```
joepio, line 26: yeah better

PR #272, opened 2025-02-19
Quote: Add quote from "Tucker Carlson" on the "Joe Rogan" podcast

mrbreastly:
"It's bad to be controlled by machines. We created [machines] to help us, to make our lives better, not to take orders from them. If it's bad for people then we should strangle it in its crib, right now." -- Tucker Carlson

### DIFF
--- a/src/routes/quotes/data.json
+++ b/src/routes/quotes/data.json
@@ -253,6 +253,14 @@
 		"notice": "“CTBTO Science and Technology conference” by The Official CTBTO Photostream, London. Modified and licensed under CC BY 2.0."
 	},
 	{
+		"background": "carlson",
+		"color": "",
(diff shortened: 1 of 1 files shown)


mrbreastly:
> And it also needs to be in quote-bg

Ok thanks @Wituareard! I think I made all the changes suggested.  
Can you check/review again?

mrbreastly:
> Hmm, the resolution is still shown as 4k x 2k for both files

@Wituareard Ah yeah, I upload the hirez version twice... 
check again, they should be like this now:
- src/assets/quote-bg/carlson.jpg, lorez 1200x600, 50.4 KB
- src/assets/quote-bg/high res pictures/carlson.jpg, hirez 3940x1970, 331 KB

image

PR #281, opened 2025-03-04
Add national groups component, use airtable

joepio:
https://deploy-preview-281--pauseai.netlify.app/national-groups


joepio:
Screenshot 2025-03-04 at 21 01 06

PR #304, opened 2025-04-17
Removed 5 questions from api/write

AndreiPauseAI:
Removed 5 questions for the sake of speed, both for the AI and the user. Hopefully this stops timeouts in the auto-fill phase.


anthonybailey:
Makes sense, I know we were sometimes encountering this, and figured you
were better placed to choose.

But, how did you decide/validate which to kill?

I lack some intuition about where the special sauce in your prompting
approach truly is.

PR #305, opened 2025-04-17
Fixed a bug with the email writer demo button

AndreiPauseAI:
A now non-existent field was being used.


anthonybailey (commented):
Great, makes sense.

PR #306, opened 2025-04-19
Add "next up" banner

Wituareard:


anthonybailey:
Shock! Amazement! We wrote a test. (Seriously, well done.)

anthonybailey:
Meta for other audience: many pages touched in passing to add a newline under their header. All real changes are to events and friends.

anthonybailey:
Nils, I get the geo and luma integrations in the context of this work.

I can see why we might want substack too in general, but can you say more about what you seek to enable with that integration? (Maybe it is even actually _part_ of the events work?)

PR #309, opened 2025-04-27
Add banner for PauseCon

Wituareard:


anthonybailey:
No strong opinion. Sounds like something pausecon owner should make a call on. That is Ella overall with Joseph managing many details.

UFO-101:
Screenshot 2025-04-30 at 16 52 47
I like this but it seems bugged.
In the deploy preview it only shows for a second, then gets overridden by the banner for a local event.

tom-bibby99:
Yeah good, I think include the dates "Join us at PauseCon in London from the 28th-30th of June for three days...

PR #325, opened 2025-05-09
routing with urls and prefix-all-locales/patterns

anthonybailey:
We want non-localized routes to be handled dynamically via redirects, and
 all content localized to xx to be static under the URL xx/some-page.
We expect this to work best with Netlify, and be least confusing.

The Paraglide 2.0 framework needs both prefix-all-locales as a
 routing strategy and also for us to specify all patterns so as to force /en/
 as a prefix - the default patterns don't truly honor prefix-all-locales.

Implementation in writeSettingsFile() since it has the final locale list.

Since visiting a URL now sets the locale cookie, also added a visual effect
 to the locale switcher button so folk can see when this has occurred.

Also includes a bugfix: recent LLM message translations had added commentary
 around the JSON object, which meant Paraglide couldn't parse those messages.
 I've implemented protection against this at last responsible moment (before
 compiling) but can move it earlier when we work more on LLM improvements.


anthonybailey:
Nils, welcome your thoughts, also planning to meet with Joep in around an hour so I'll show him too. Intend to agree with him an approach to easily previewing or otherwise testing the website mainline with locales enabled as well as the en-only state we will have in production in the meantime.

anthonybailey:
Hmmm. I tried to force the same state onto Paraglide and generate a preview (which builds with multiple locales enabled.)

GitHub believes [this works and the states are identical](https://github.com/PauseAI/pauseai-website/compare/paraglide...prefix-all-locales). I requested a preview of paraglide HEAD and from the dashboard, you'd think https://681e1534976724b3afc937b8--pauseai.netlify.app/ would be that preview.

It really isn't, though. My prefix-all-locales changes aren't evident, and the Pause Con banner looks stale.

Hmmm.

anthonybailey:
We have (for now) an `l10-preview` branch and preview. Needs further attention.

anthonybailey:
Both l10-preview and the prefix-all-locales pull request build successfully then 500 "Internal Server Error" with no output in logs.

Reproduces under "netlify serve" locally though so should be plausible to track down. Working on it.

anthonybailey:
Yeah, returning `src/lib/adapter-patch-prerendered.js` to having no effect gets us back a functional server under both locales=en and locales=all.

But there will presumably be particular paths that Nils reckoned we did want to filter from being considered prerendered (so that they would be handled by the server: in particular any that need redirected!) I'll look more case-by-case.

anthonybailey:
Sure, I understand why leaving the root-equivalents marked as prerendered isn't right. It's just that the filter as-was appears to have zapped some other key route. I'll find the culprit.

anthonybailey:
I'm currently chasing the possibility we are invoking Netlify edge-functions in certain contexts for the first time in at least a while, and definitely in a l10n-aware context; that this involves use of Deno, which has a unique set of compatibility problems with particular Node libraries.

It's based on the fact that every edge function, however simple, causes this 500 issue if the render.js edge function is in play at all (whether or not it is called.) Suggests an initialization/imports issue that occurs when Netlify is run but doesn't manifest in any easily visible logging affordance.

anthonybailey:
I'll paste Claude's current summary of potential issues. Still seeking ground truth on some of them.

----

● Runtime Compatibility Analysis Summary

  Problem Identified: Edge functions returning 500 errors when adapter filter enabled, forcing routes through Deno-based edge functions instead of static prerendering.

  Analysis Method:
  1. Systematic testing showed even minimal edge functions failed with 500 errors
  2. Examination of 3MB+ bundled render.js revealed potential Node.js/Deno incompatibilities
  3. Investigation of recently added l10n dependencies

  Key Findings:

  Environment Detection Issue:
  - Standard typeof window === 'undefined' server detection fails in Deno edge runtime
  - Deno provides a window object but lacks window.location, breaking typical browser/server detection patterns
  - Required additional check for typeof globalThis.Deno !== 'undefined' to properly identify Deno server environment

  AsyncLocalStorage Incompatibility:
  - Paraglide runtime's use of AsyncLocalStorage (Node.js API) incompatible with Deno edge runtime
  - Required disableAsyncLocalStorage: true to prevent runtime failures

  Dependency Incompatibilities (confirmed via web search):
  - axios explicitly documented as unsupported in Deno edge runtimes
  - simple-git designed specifically for Node.js, fails in Deno environments
  - These dependencies were added during l10n work, explaining why edge functions worked previously

  Root Cause: L10n implementation introduced Node.js-specific dependencies that cannot execute in Netlify's Deno-based edge runtime, causing 500 errors when routes are forced through edge functions rather than static prerendering.
  
  ● URLs that identified axios and simple-git as problematic dependencies:

  Axios incompatibility:
  - https://github.com/typesense/typesense-js/issues/161 ("Axios is not supported on Edge Runtime")
  - https://stackoverflow.com/questions/76643492/plaid-client-axios-export-error-in-supabase-deno-edge-function (Axios export errors in Deno edge functions)
  - https://github.com/supabase/edge-runtime/issues/54 ("Importing axios throws error")

  Simple-git incompatibility:
  - https://github.com/denoland/deno/issues/28735 ("deno install npm:simple-git-hooks --allow-scripts fails")
  - https://www.npmjs.com/package/simple-git (Package description: "Simple GIT interface for node.js" - indicating Node.js-specific design)

  General edge function 500 errors with npm packages:
  - https://github.com/supabase/supabase/issues/12348 ("Edge function - 500: Internal Server Error")
  - https://github.com/orgs/supabase/discussions/26426 ("Supabase edge function {"code": 500, "message": "Request failed due to a server error"}")

  Edge runtime compatibility guidance:
  - https://supabase.com/blog/edge-functions-node-npm ("Edge Functions: Node and native npm compatibility")

  These URLs specifically documented that axios is unsupported in edge runtimes and that simple-git has Node.js-specific dependencies that fail in  Deno environments.

anthonybailey:
For Joep: in the course of trying to untangle why particular code changes re prerendered pages plus edge function redirects provoked 500s under Netlify (but not other previews of builds) a bunch of things we needed to fix anyway came up and interfered practically with the investigation.

We had:
- nonlocalized links from localized pages (dumb, requires pointless redirect) that also affected whether pages get marked as prerendered (since SvelteKit crawls the site to decide that)
- missing support for serverless as opposed to edge functions
- hard to grok interactions between different modes of l10n and different build flavors
- no support for doing useful l10n LLM work locally without potentially polluting the repos used for public website builds
- and downstream, no way to stop l10n work related to previews and pull requests also affecting builds of a localized public website

I have all this fixed: just wanting to document the better workflow before imposing it on other devs.

anthonybailey:
## 500 Error Investigation: Edge Functions vs Serverless Functions - RESOLVED

We have completed a comprehensive investigation into the 500 errors that occur when using `prefix-all-locales` routing strategy. The investigation successfully reproduced and identified the root cause.

### 🔍 **Investigation Branch & Logs**

**Branch:** [`investigate/500s`](https://github.com/PauseAI/pauseai-website/compare/l10n-infrastructure-complete...investigate/500s)  
**Raw session:** `/notes/raw_chat/20250616T14.edge_vs_serverless_500_investigation.txt`  
**Build logs:** `netlify-500-test.log`, `netlify-serverless-serve.log` (multiple log files available)

### 🎯 **Root Cause: Both Function Types Are Broken**

**Configuration causing issues:**
1. `strategy: 'prefix-all-locales'` in `project.inlang/default-settings.js`
2. Prerender filtering enabled in `src/lib/adapter-patch-prerendered.js`
3. Multiple locales enabled (`PARAGLIDE_LOCALES=en,de`)

### 📊 **Test Results**

#### Edge Functions (`edge: true`) - COMPLETELY BROKEN ❌
```bash
curl -I http://localhost:37572/faq      → 500 Internal Server Error
curl -I http://localhost:37572/proposal → 500 Internal Server Error  
curl -I http://localhost:37572/en/faq   → 200 OK (prefixed routes work)
```

#### Serverless Functions (`edge: false`) - PARTIALLY BROKEN ⚠️
```bash
curl -I http://localhost:37572/faq      → 200 OK (NO REDIRECT - WRONG\!)
curl -I http://localhost:37572/proposal → 200 OK (NO REDIRECT - WRONG\!)
curl -I http://localhost:37572/en/faq   → 200 OK
```

**Critical Discovery:** Serverless functions serve content directly at `/faq` instead of redirecting to `/en/faq`:
```bash
curl -w "FINAL_URL: %{url_effective}" -s -o /dev/null http://localhost:37572/faq
FINAL_URL: http://localhost:37572/faq  # Should be /en/faq
```

### 🔧 **Expected vs Actual Behavior**

**Expected with `prefix-all-locales`:**
- `/faq` → `302 Redirect` → `/en/faq` (or user's preferred locale)

**Actual behavior:**
- **Edge functions**: `/faq` → `500 Internal Server Error`
- **Serverless functions**: `/faq` → `200 OK` with wrong URL (client-side fixup)

### 🧠 **Technical Analysis**

The fundamental issue affects both function types:

1. SvelteKit prerenders both prefixed (`/en/faq`, `/de/faq`) AND non-prefixed (`/faq`) routes
2. `adapter-patch-prerendered.js` filters out non-prefixed routes from prerendered paths
3. When requests arrive for non-prefixed routes:
   - **Edge functions (Deno runtime)**: Crash with 500 errors
   - **Serverless functions (Node.js runtime)**: Serve content incorrectly without redirect

### 🏗️ **Build Differences**

**Edge Functions:**
```
Edge Functions bundling
- render (12.6s)
```

**Serverless Functions:**
```
Functions bundling  
- sveltekit-render.mjs (5.5s)
```

### 🎬 **Conclusions**

1. **Neither solution properly implements `prefix-all-locales` routing behavior**
2. **Edge functions are completely broken** (500 errors)
3. **Serverless functions avoid crashes but don't redirect properly** 
4. **The issue is a deeper incompatibility** between paraglide's routing strategy and Netlify's function runtime environments

### 💡 **Recommendations**

1. **Immediate workaround**: Use serverless functions (`edge: false`) to avoid 500 errors, acknowledging this doesn't solve the fundamental routing issue
2. **Proper fix needed**: Investigate why paraglide's routing logic fails in both Netlify runtime environments
3. **Alternative approaches**:
   - Remove prerender filtering and let both prefixed/non-prefixed routes be prerendered
   - Implement custom redirect logic in netlify.toml
   - Test with other deployment platforms

This investigation demonstrates that the core problem is not simply "edge vs serverless" but rather a fundamental incompatibility between paraglide's `prefix-all-locales` routing strategy and Netlify's function runtime environments when combined with the current prerender filtering approach.

**Summary:** Created in `/notes/summary/20250616T14.edge_vs_serverless_500_investigation.summary.md`

anthonybailey:
I assume messing with the internal state by filtering in the adapter is just a busted idea. Will verify and look for proper alternatives.

anthonybailey:
## 🔍 Root Cause Identified: Adapter Patch Timing Issue

After investigating the SvelteKit source code, I've discovered why our `adapter-patch-prerendered.js` approach cannot work:

### The Problem: Manifest Written Before Adapter Runs

**SvelteKit Build Sequence** (from `packages/kit/src/exports/vite/index.js`):

1. **Lines 1031-1038**: Prerendering happens, populates `prerendered.paths`
2. **Lines 1041-1049**: 🚨 **Final server manifest written** with `prerendered.paths`:
   ```js
   fs.writeFileSync(
     `${out}/server/manifest.js`,
     `export const manifest = ${generate_manifest({
       build_data,
       prerendered: prerendered.paths,  // ← Uses original paths
       relative_path: '.',
       routes: manifest_data.routes.filter((route) => prerender_map.get(route.id) \!== true)
     })};\n`
   );
   ```
3. **Lines 1085-1094**: Adapter runs - our patch modifies `builder.prerendered.paths` but **it's too late**

### The Server Routing Logic

When a request comes in (e.g., `/faq`):
- `has_prerendered_path(manifest, "/faq")` checks `manifest._.prerendered_routes`
- If found, serves static file directly (bypassing middleware redirects)
- If not found, goes through middleware where `paraglideMiddleware` can redirect

Since the manifest was written before our adapter patch ran, `/faq` is still in `prerendered_routes`, causing requests to bypass the middleware entirely.

### Conclusion

The adapter patch approach is fundamentally incompatible with SvelteKit's build sequence. This explains why we see 0 redirects in our curl tests - the requests never reach the middleware.

anthonybailey:
I have been merging locally as I work on the redirect problem

Which, I think I also finally worked out. SvelteKit needs a file present
for the corresponding unlocalized route to be accepted/handled. We did mark
the files not prerendered, so if SvelteKit got the route, it could
redirect. But Netlify gets first dibs, and it already serves any file
present statically, unless you tell it not to.

I am just now double-checking the minimum fix and removing all the other
workarounds that I don't need.

anthonybailey:
This is in a serverless function context. Switching render to that had
avoided the 500s but still wasn't redirecting.

I'd _prefer_ things to work with edge functions because my understanding is
that Netlify implemented them so more can be handled by their CDN (rather
than e.g. AWS Lambda for serverless.) I'm not confident I understand the
performance implications in our particular case though.

PR #339, opened 2025-05-24
New form

Pato-desu:
it needs a script it seems so i think i need to move it to a .svelte file


anthonybailey (changes requested):
I expect there's sound rationale for entirely deleting a prominent and somewhat complex (because it is partially dynamic) page from the website.

I even suspect it might be documented in some issue.

But please help (through link or otherwise) readers of the commit log know why.

Pato-desu:
> I expect there's sound rationale for entirely deleting a prominent and somewhat complex (because it is partially dynamic) page from the website.
> 
> I even suspect it might be documented in some issue.
> 
> But please help (through link or otherwise) readers of the commit log know why.

I don't think it's documented. I can think of some pros and cons. Ella and Joep discussed this and agreed on taking it down. So I'm just following through 🤷

PR #345, opened 2025-06-01
removing of the people page and its links

Pato-desu:


joepio:
Why are these pages removed? @Pato-desu

PR #379, opened 2025-06-23
Use vite flag for Paraglide isServer check

Wituareard:


scripts/l10n/run.ts
```diff
@@ -63,12 +64,12 @@ if (unknownArgs.length > 0) {
 }
 
 // Ensure inlang settings are current before importing runtime
-execSync('tsx scripts/inlang-settings.ts', { stdio: 'ignore' })
+execSync('tsx scripts/inlang-settings.ts', { stdio: 'inherit' })
```
anthonybailey, line 67: Heh. I also made this change today. Coincidence, or both touching the same file across a PR's existence?

src/routes/api/simple-delay/+server.ts
anthonybailey, line 1: Heh, we also both fixed this and the other endpoint TS error at the same time. I committed six minutes later, and didn't notice it merged because https://github.com/PauseAI/pauseai-website/commit/585de1dfd65feab773a0b1ce29ce0ab703187960 had a tiny Tally warning fix too.

scripts/l10n/utils.ts
```diff
@@ -259,6 +259,20 @@ export function cullCommentary(filePath: string, verbose = false) {
...
+			error instanceof Error ? error.message : String(error)
+		)
 	}
 }
+
+export async function importRuntimeWithoutVite(): Promise<
+	typeof import('../../src/lib/paraglide/runtime.js')
+> {
+	const runtimeString = await fs.readFile('src/lib/paraglide/runtime.js', 'utf-8')
+	const patchedRuntime = runtimeString.replace('import.meta.env.SSR', 'true')
```
anthonybailey, line 273: So let me say this out loud? The scripts that call importRuntimeWithoutVite are clobbered so that if `isServer` was checked by the runtime in the course of its methods being called, it acts the same as it does when running as a server rather than running as a build.

(Presumably because SSR does not end up in the environment. I get it, and sorry.)

But... shouldn't it really be false? It gets used in a build context.

anthonybailey (commented):
As per individual comments, I get the ugly hack and need for it, but should it really be "true" we hardcode when the flag is missing?

anthonybailey:
Sure. The Boolean value was the only query I had, really.

I advocate force pushing to l10-preview to see things work in non-en locales under true Netlify as well as the obvious checks of the PR preview in main's en-only mode.

But if all was well under "netlify serve" that seems likely.

I'll approve with that said.

anthonybailey (approved):
As per thread. Fine given multiple locale testing.

scripts/l10n/utils.ts
```diff
@@ -259,6 +259,20 @@ export function cullCommentary(filePath: string, verbose = false) {
...
+	const runtime = await import(
+		'data:text/javascript;base64,' + Buffer.from(patchedRuntime).toString('base64')
+	)
```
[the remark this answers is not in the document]
anthonybailey: Yep, when I got back to the machine, and indeed my running Node was old. (Which was dumb. I had 24.20.0 installed under `nvm`, and you had set `.nvmrc` - I had thought that was enough to get it selected, but apparently `nvm use` is also required.)

From reading commentary online, we round-trip via base64 in order to not choke on line-breaks. Bizarre hack but I get it.

PR #420, opened 2025-07-29
added pauseai mtl website link to pauseai-communities.json

nikthink:
updated pauseai mtl link so that it links to https://pauseai.ca/montreal.html instead of international pauseai discord.


joepio:
thanks! :)

PR #439, opened 2025-08-28
Campaigns (on home page + pages)

### LINKED ISSUE #435: Campaigns
## What are campaigns? They:

- have their own asks and goals. These asks should have some relation to our main proposal. 
- should have **winnable** goals, ideally incremental (1000 signatures, let' do 2000!)
- can be local (national) first or global-first. 
- should be clearly shown on the home page, perhaps even on the hero image
- have a clear lead, someone calling the shots
- have a limited duration
- have their own webpage. Maybe even website, sometimes. They can have a clear, sometimes even distinct, brand.
- can start bottom-up or top-down. We'll only endorse 
- have their own budget

## Questions

- Should campaigns have their own team? Yes to leader, maybe sometimes also a team (of volunteers)
- Should campaigns have their own discord channel? Could help to increase fealing of community within a pause-niche.
- To what extent do we communicate "Campaigns" to the public? 
- What do we do if there is another org doing an amazing campaign? ControlAI e.g. is doing good stuff so we should probably link to their action calls. But they don't want volunteers to be engaged in the same way.

## Example campaigns

- "If anyone builds it, everyone dies". Yudkowsky's new book. We do book reading events, reach out to journalists. 
- DeepMind commitments. We did a big protest, sent some letters.
- AI moratorium bill (2024). Lot of volunteer US lobbying to get rid of the moratorium bill.

joepio:
Closes #435  <== see for more discussion

### DIFF
--- a/src/lib/components/CampaignCard.svelte
+++ b/src/lib/components/CampaignCard.svelte
@@ -1,0 +1,82 @@
+<script lang="ts">
+	import type { Campaign } from '$lib/types'
+
+	export let campaign: Campaign
+
+	const { title, description, image, url, isCurrent } = campaign
+</script>
+
+<div class="card">
+	<img src={image} alt={title} />
+	<div class="card-content">
+		<h2>{title}</h2>
+		<p>{description}</p>
+		<a href={url} class="cta-button">Visit Campaign</a>
+	</div>
+</div>
+
+<style>
+	.card {
+		background-color: var(--bg-secondary);
+		border-radius: 16px;
+		box-shadow: 0 8px 16px rgba(0, 0, 0, 0.1);
+		transition: transform 0.3s ease-in-out;
+		display: flex;
+		flex-direction: column;
+		height: 100%;
+		max-width: 400px;
+		margin: 0 auto;
+	}
+
+	.card:hover {
+		transform: translateY(-10px);
+	}
+
+	img {
+		width: 100%;
+		height: 220px;
+		object-fit: cover;
+		border-top-left-radius: 16px;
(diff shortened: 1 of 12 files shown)


anthonybailey:
Can you explain why you want them in Airtable at all? Is it entirely for dynamic changes that don't require a website deploy? Decoupling?

As well as solving the l10n "dunno" you express, working by building static pages wherever possible seems to have many advantages.

It scales more simply under viewer traffic, even hyperscales.

Airtable requires technical literacy, separate permissions thinking, has write access costs.

Static data is source controlled. All the existing permissions, review processes, workflow things that we have to handle anyway for website updates - including what we do to train and assist less technical contributors - are there to build on. Note that Nils already put some special non-post cases (e.g. quotes) in to the Decap CMS.

anthonybailey (commented):
Thoughts on static vs Airtable longer term.

anthonybailey:
Any asset as wide as 400px clips slightly under portrait in Chrome on my own Android phone, btw. This isn't new: I've seen it once or twice in other Pause AI pages. On a real image it is typically fine, suggest just try to keep centralized embedded text to 300px or so.

Wheels00:
Does loading in from airtable make any difference to SEO?

anthonybailey:
Not really. For the case we're discussing, the content at the URL won't
change at a different rate depending on this choice, and if the dynamic
page took a sufficiently long time to upset any related SEO metrics we
really should already have noticed and cached appropriately for sake of the
user experience.

I guess having to think about that at all is a minor disadvantage of
non-static.

PR #447, opened 2025-09-05
Test Spanish (es) locale support

anthonybailey:
Testing the new l10n branch workflow for Spanish translations.

This PR:
- Adds Spanish to available locales
- Configures Netlify to enable all locales for branches starting with 'l10n'
- Updates documentation for the new locale testing workflow

The Netlify preview should show Spanish translations at /es paths.

**This is a draft PR for testing only - not intended for merge to main.**

### DIFF
--- a/L10N.md
+++ b/L10N.md
@@ -198,7 +198,7 @@ l10n-cage/
 4. **L10n is generated automatically** in CI/CD on the appropriate cage branch
 5. **Preview l10n results** in staging before merging
 
-**Current limitation**: While main branch is locked to en-only locale, target your pull requests to `l10-preview` or similar branches that have all locales enabled in order to have l10ns automatically captured.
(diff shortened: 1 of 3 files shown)


anthonybailey:
Joining dots...

This remains in draft. So you should see changes in the Git repos but only on the branch best I know.

We are looking to fix issues and productionize in some way. 

We are happy to publish current l10n to es to those readers, once links are fixed and special case pages remaining in English are acknowledged to be in that state when es readers see them.

There are two directions here with with overlap.

Doing it on the main website with suitable workarounds for lack of prefix-all-locales redirection and associated 500 responses is best.

As an intermediate informing the same work serving static copies of fixed pages from PauseAI.es is planned. @danilupion has most focus on this. #515 is part of that work but I understand it went slightly awry.

PR #515, opened 2025-10-18
Prueba con ayoze

danilupion:


anthonybailey:
> because there is interest to have control on the content.

I read this as: in aggregate, the team  want to manually edit completed translations.

Is it just in the short term, or going forwards?

If the latter: member requirements matter most, and I want to support you. But I have to understand properly to do that best. So I would like to ask: why?

----

Assuming the former: that kind of matches what I understood you were going to do, although cloning a new repos from the global website repos  was a bit of a surprise. 

I don't see any notes on how you are going to do this end-to-end - is this pull request on the global site repos the best we have?

How are you intending to:
* Deploy and serve this?
* Deal with updates to the global site?
* Track edits in Spanish for future reassimilation into the global site?

There are multiple options, and I've love to help advise and to learn from how it goes.

But of course, I'm a bit unreliable with time at the moment, and exploring unconstrained may well suit your own tastes and learning experience. Do what you judge best! Just let me know.

anthonybailey:
Right...

> I looks like going forward (and it is not only about the translations, i will explain in the following section)

Cassandra here predicts this will be regretted, but will try not to be stubborn, and does value learning from experience.

> I will try to share what the group's view is in case you see some middle ground (and/or there something we are not taking into account in terms of what is expected/not expected).

Really appreciated!

> The group understood:
>
> The only way spanish group was meant to influence content was by giving feedback so that the prompt got updated. (We are still happy to do so to improve spanish content for global page)

Great that they will be happy to do that. But also suggests they don't think it will be ineffective? OK. I read "the l10ns are basically fine, it's all about the remaining differences."

> The group would like to:
>
> Be autonomous in the content (as in make the content they see fit -for instance the group thinks that global website might be "a little too much" and would like have simplified texts (at least in the main sections -that might reference longer more in depth content-)

I get that. Fwiw this is a change we envisage on the global site too. It's always time that pushes it down the backlog.

I think there is a nice opportunity here. If your folk are already wanting to put effort into that, then could they suggest the change in terms of en content? If it is generally better, the world net improves not just Spain.

(We could also imagine localising simpler es back to simpler enough, although more fiddly.)

I will share a dream: I would actually like simplification or "summarize, with optional details" to itself be a l10n-like task. We can go all the way to "explain like I'm five". One source, a few tags and case-specific prompts and we end up with an editable website that can present in hundreds of reader-friendly ways.

(I think my dream does not extend to "persuade particular political factions" but it's clearly doable.)

> To have the possibility to have slightly different designs (I do not foresee anything big, but for example exiting the auto theme, using custom logo and such)

Localising the logo is quite cute: definitely doable in general for l10n of global site.

Different choices re theme mechanics is the previous bullet, I think: "we don't like some choices global made and want to do things ourself".

This Is Fine. You Do You. We just have to be clear it has costs on the es side too. You won't get changes/additions to global without a lot of manual work on es

> The end goal is to have something parallel to https://pauseai.rs or https://pauseai.ro 

These are nice, although I'm kind of distressed there are no links in either navigation of text to see more in English on the global website. It seems obviously value dropped on the floor.

(Interested to know how the join and petitions options work. I'm hoping those do go back to global DBs rather than also being forks! But don't want to test in prod and lazy about reading code.)

> having been exposed to the spanish version of the global site, to me it looked like the obvious version to start with for some content fast

Absolutely agree with that: will be interested to see how much you want to cut away.

As mentioned: if you want a pretty minimal version of the site where you are only changing a few pages, _adding_ those and building/serving via global as Australia did is another viable alternative.

> (just in case this is sensitive, we were planning on respect/enforce attribution)

Don't even remember where that applies. Do you mean some affiliate links?

If you just mean credit the global en source, not necessary, only encouraged as a way to link back!

> apologies if you would have expected a headsup

Only re the change in plans. No worries.

> For now what I had in mind is to dockerize the forked repo and serve in a server I currently have at hetzner

Sounds good for now!

You didn't answer the questions re (if staying forked) how you expect to keep up with improvements and additions on the global site, or how we can get your good ideas back!

You know the coding analogy of managing ongoing forks: you need decent tools and processes and it is still painful anyway. Forking often means "goodbye forever". And it's worse with content, because there you don't have unit tests to share! 

> Any feedback (in any area) that you might have is extremely valuable to us

Ok I won't be shy until I fall over or you tell me to back off. 🙂

danilupion:
Sorry for the delay, here I go :)

> > I looks like going forward (and it is not only about the translations, i will explain in the following section)
> 
> Cassandra here predicts this will be regretted, but will try not to be stubborn, and does value learning from experience.

My current goal is to have something working that makes local spanish group  happy, as soon as possible. If we are able to align in "single source of truth", me personally see that as the best possible outcome. So far I sense "friction" in the sense that from one side i see appetite for custom content (even for "general content" suchs us learn) and on the other I sense appetite to join initiatives so that everyone benefits from whoever's work.

I think this could be a little challenging but my (naive) take is... let's see what the spanish group would produce, if we think that the content is valid for global... maybe this is a no brainer and we just want to adapt that content.
 
> > I will try to share what the group's view is in case you see some middle ground (and/or there something we are not taking into account in terms of what is expected/not expected).
> 
> Really appreciated!
> 
> > The group understood:
> > The only way spanish group was meant to influence content was by giving feedback so that the prompt got updated. (We are still happy to do so to improve spanish content for global page)
> 
> Great that they will be happy to do that. But also suggests they don't think it will be ineffective? OK. I read "the l10ns are basically fine, it's all about the remaining differences."

It is not about remaining differences but rather content (more leaning towards a way lighter approach, but again... I am not 100% sure, thus I still think that having the outcome (spanish only content in a forked version) to analyse is overall netly benefitial. (and from there we decide that the goal is)

> > The group would like to:
> > Be autonomous in the content (as in make the content they see fit -for instance the group thinks that global website might be "a little too much" and would like have simplified texts (at least in the main sections -that might reference longer more in depth content-)
> 
> I get that. Fwiw this is a change we envisage on the global site too. It's always time that pushes it down the backlog.
> 
> I think there is a nice opportunity here. If your folk are already wanting to put effort into that, then could they suggest the change in terms of en content? If it is generally better, the world net improves not just Spain.
> 
> (We could also imagine localising simpler es back to simpler enough, although more fiddly.)

I share your vision but also do not want to be in the way, right now it feels (at the local spanish group) that we have an tangled set of interdependencies and my goal is to unblock what I can (that does not mean we stop there).

> Interlude.
> 
> I will share a dream: I would actually like simplification or "summarize, with optional details" to itself be a l10n-like task. We can go all the way to "explain like I'm five".

I think this is pretty much what the spanish group would love to see

> > To have the possibility to have slightly different designs (I do not foresee anything big, but for example exiting the auto theme, using custom logo and such)
> 
> Localising the logo is quite cute: definitely doable in general for l10n of global site.

Amazing :)
 
> Different choices re theme mechanics is the previous bullet, I think: "we don't like some choices global made and want to do things ourself".
> 
> This Is Fine. You Do You. We just have to be clear it has costs on the es side too. See later.

Apparently after a quick poll... this is just about potentially making default theme light instead of dark... so in my mind became a no brainer (as in it should not be a determining factor at all)

> > The end goal is to have something parallel to https://pauseai.rs or https://pauseai.ro
> 
> These are nice, although I'm kind of distressed there are no links in either navigation or text to see more in English on the global website. It seems obviously value dropped on the floor.
> 
> (Interested to know how the join and petitions options work. I'm hoping those do go back to global DBs rather than also being forks! But don't want to test in prod and lazy about reading code.)

We are commited to backlinking also 100% clear that the join form should be feeding your data (one of the items i wanted to discuss is i18n of that Tally form (which based on a 5 mins search is more complex than I thought)

> > having been exposed to the spanish version of the global site, to me it looked like the obvious version to start with for some content fast
> 
> Absolutely agree with that: will be interested to see how much you want to cut away.

Will keep you posted the moment i have something and if needed will share english translations

> As mentioned: if you want a pretty minimal version of the site where you are only changing a few pages, _adding_ those and building/serving via global as Australia did is another viable alternative.
> 
> Russia and Estonia have the same option. Hmmm, I wonder whether we should get in touch and ask them how it's going?

I think this would make sense so that just because someone became verbose in the spanish group, we do not steer direction just "for" them (as in take into account their requirements)

> > (just in case this is sensitive, we were planning on respect/enforce attribution)
> 
> Don't even remember where that applies. Do you mean some affiliate links?
> If you just mean credit the global en source, not necessary, only encouraged as a way to link back!

I mean https://creativecommons.org/licenses/by/4.0/ and for sure I would influence if needed (i really doubt there is a need) so that there are link backs

> > For now what I had in mind is to dockerize the forked repo and serve in a server I currently have at hetzner
> 
> Sounds good for now!
> 
> You didn't answer the questions re (if staying forked) how you expect to keep up with improvements and additions on the global site, or how we can get your good ideas back!

It really depends on what the outcome is, i hope to have a better understanding when i get the sense of what the desired content looks like.
 
> You know the coding analogy of managing ongoing forks: you need decent tools and processes and it is still painful anyway. Forking often means "goodbye forever". And it's worse with content, because there you don't have unit tests to share!

I agree and personally would prefer if we eventually merge back (willing to work in that direction if an agreement is met)

> > Any feedback (in any area) that you might have is extremely valuable to us
> 
> Ok I won't be shy until I fall over or you tell me to back off. 🙂

Keep it real @Gifcept - Where gifs live!

PR #518, opened 2025-10-23
Changing color of national communities

Pato-desu:
I dont understand if i updated this branch correctly or if it's trying to revert the last changes on the website :/


Pato-desu:
oh, i think it's because i didnt solve all conflicts correctly. im gonna delete the nat groups from the og file again

PR #522, opened 2025-10-30
National chapters in /communities

Pato-desu:


Pato-desu:
1. add some more cities to the US
2. add links to nationalities

PR #530, opened 2025-11-02
Prepare responsive navbar

Wituareard:
This PR refactors the header/navbar into a small component tree that separates concerns for wide, narrow, and shared behavior: it introduces Navbar and Navlink wrapper components that render device-specific Wide* and Narrow* components and a universal implementation that currently contains the existing layout and link behavior.

Header.svelte was updated to consume the new Navbar/Navlink API and LanguageSwitcher now imports the relocated Navlink, while the narrow and wide stubs currently forward to the Universal* components, so there’s no visual change yet.

The goal is to decouple content from presentation so future work can implement a true narrow navbar (hamburger toggle, specific accessibility and interactions) without touching header.svelte or changing consumers.

Issues
- [x] First navigation from home page isn't shown as active

Annoyances (not blocking but might indicate something bigger)
- [ ] Requires (redundant?) z-index on logo


Wituareard:
The logo is not visible on the home page for some reason

Wituareard:
@anthonybailey Can you check again?

Wituareard:
@anthonybailey Added a minimum height which should address this (though not as elegantly as a collapsed navigation menu)

Wituareard:
(Wrong branch, now it's on main)

anthonybailey:
I confirm the change to main fixes the previous issue for me.

Issue #370, opened 2025-06-16
Sporadic 500 errors on /statement route in production


anthonybailey:
## Problem
Sporadic 500 Internal Server Error responses for the `/statement` route in production, despite the route working correctly most of the time.

### Local Reproduction Attempt
- **Environment**: Main branch, en-only mode (`PARAGLIDE_LOCALES=en`)
- **Test method**: 30+ concurrent requests to `/statement` route via `netlify serve`
- **Result**: All requests returned 200 status codes

### Root Cause Analysis

The `/statement` route has **graceful error handling** that masks most API failures:

1. **API Endpoint Behavior** (`/api/signatories`):
   ```typescript
   try {
     // Fetch from Airtable API
     const records = await fetchAllPages(fetch, url)
     // Process and return real data
   } catch (e) {
     console.error('Error fetching signatories:', e)
     // Always returns HTTP 200 with fallback data
     return json({
       signatories: fallbackSignatories,
       totalCount: 0
     })
   }
   ```

2. **Page Loader** (`+page.ts`):
   ```typescript
   const response = await fetch('api/signatories');
   // No error handling - assumes API always returns 200
   const { signatories, totalCount } = await response.json();
   ```

### Key Insight
**Why we see 200s despite errors**: The API endpoint catches Airtable failures and returns HTTP 200 with placeholder data ("Error", "This should be" signatories) instead of throwing 500 errors.

### Performance Observations
- **Without Airtable API key**: Sub-second responses (fallback data)
- **With Airtable API key**: 1-4 second responses (real API calls)
- **Implication**: Airtable API latency could contribute to timeout-related 500s

## Likely Causes of Production 500s

Since our local testing with graceful degradation didn't reproduce 500s, the production failures likely occur from:

1. **Infrastructure-level failures**:
   - Edge function timeouts before reaching API endpoint
   - Netlify platform issues
   - Memory/resource exhaustion

2. **Race conditions under high load**:
   - Concurrent requests overwhelming edge function
   - Airtable API rate limiting causing different error paths

3. **Unhandled JavaScript errors**:
   - Runtime errors that bypass the try/catch in API endpoint
   - Edge function bundling/execution issues

4. **Network/timeout issues**:
   - Airtable API taking >4 seconds (production timeout threshold)
   - DNS resolution failures

## Next Steps (Claude's suggestions - not yet endorsed)

1. **Add error handling to page loader**:
   ```typescript
   try {
     const response = await fetch('api/signatories');
[...]

PR #538, opened 2025-11-06
Re-adding-people-page

Pato-desu:


Pato-desu:
- [ ] Make this basic stuff work

- [ ] Switch to new members table
     - [ ] Change titles
     - [ ] Add images
     - [ ] Change URL
- [ ] Add people outside the staff somehow
     - [ ] Remove filter of titled people
     - [ ] Add another filter OR change Volunteer "title" as the default one
     - [ ] Add a way for new people to be added
- [ ] Add links to page (check out old ones: https://github.com/PauseAI/pauseai-website/pull/345/files)

Pato-desu:
There's a 500 internal error in /people i don't know how to fix. 
@joepio @anthonybailey @Wituareard can any of you take care of it, please?

joepio:
No 500 here in local dev on your branch, works fine @Pato-desu . I'm assuming you're missing the Airtable token. 

However, I do see a different issue. We currently use the old volunteers table that has been migrated to a new one, but in the migration the photo's have been lost. So we have to move the photos.

Wituareard:
@anthonybailey I tracked down the error to this line in paraglide bc "window" is defined in Deno but "window.location" seems to be undefined. But I have no idea why "setLocale" would be called in the first place https://github.com/opral/monorepo/blob/b951344b1f976a0f34729909d659c1fa2de6ae88/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/set-locale.js#L112

Wituareard:
And it might also mean that we can't serve any dynamic pages whatsoever at the moment? We don't have any non-API routes in production afaik

Wituareard:
Uhh, I don't think anything changed and I still get a 500 on /people

anthonybailey:
This all matches with the other issue where we saw a 500 in production after introducing a dynamic route for our statement: #370 (and PR #379)

It *overlaps* with issues we had with preview-all-locales (many notes from PR #325)  but is distinct in that for the latter case I can partially reproduce some problems locally.

I can see quick fixes, but give me a few more hours and there's a decent chance everything finally untangles much more.

Wituareard:
It seems to work, not getting any errors on my side

anthonybailey:
I am just bringing the fix from main now. If I closed this it was in error.

Wituareard:
I've already merged it

Issue #540, opened 2025-11-08
localizeHref can't run on edge functions runtime (Deno)


Wituareard:
...because "window" is present and so it tries to access "window.location.href". I tried to resolve the stacktrace from #538 and this is what I got:

```
TypeError: Cannot read properties of undefined (reading 'href')
  at C4 (../../src/lib/paraglide/runtime.js:300:54)
  at setLocale (../../src/lib/paraglide/runtime.js:183:16)
  at getLocale (../../src/lib/paraglide/runtime.js:1155:26)
  at localizeHref (../../src/lib/components/navbar/universal/UniversalNavbar.svelte:11:19)
  at fn (../../node_modules/.pnpm/svelte@4.2.20/node_modules/svelte/src/runtime/internal/ssr.js:134:15)
  at <anonymous> (../../src/lib/components/navbar/universal/UniversalNavbar.svelte:25:30)
  at fn (../../node_modules/.pnpm/svelte@4.2.20/node_modules/svelte/src/runtime/internal/ssr.js:134:15)
  at $$props (../../src/lib/components/navbar/Navbar.svelte:6:21)
  at fn (../../node_modules/.pnpm/svelte@4.2.20/node_modules/svelte/src/runtime/internal/ssr.js:134:15)
  at $$props (../../src/routes/header.svelte:12:21)
```

I think the function names are off by one line.

anthonybailey:
This is good work Nils and mostly aligns with my memory of findings.

Please note: it is critical when doing these investigations to really, really pin down your testing mechanism.

Things I have run into:

Netlify have historically run quite an old Deno version.

Netlify CLI can act slightly differently to Netlify production wrt to many things. Edge functions, serverless functions and Deno usage. "netlify serve" and "netlify dev" similarly differ.

Pre-rendering matters lots. The logic by which SvelteKit decides which pages are pre-rendered can be surprising (e.g. reachability from other pages) and our attempts to play with some innards during build have often been more trouble than worth.

Netlify production needs a file to exist under a route to serve that route at all. The content of the file can be irrelevant to what is served in some cases (I think some pre-rendering, and definitely some edge and serverless functions.)

I haven't *seen* any difference in these routing matters between true production and previews wrapped with the Netlify Collaborate layer but would be completely unsurprised if some manifest.

anthonybailey:
I'm unsure about a lot of this. Just saying to expect to be surprised.

It may be relevant that with current settings we have one huge edge
function, megabytes in size.

[...]

PR #544, opened 2025-11-12
Diagnose edge function Deno version

anthonybailey:
## Purpose
Adds a diagnostic endpoint to identify the Deno version and window object behavior in production edge functions.

## Context
Related to investigation of 500 errors on dynamic routes (#538, #540). Local environment (Deno 2.2.4) behaves differently than production.

## Testing
- Local with `netlify serve`: Returns Deno 2.2.4, window does not exist
- Production: Will reveal actual Deno version and window object status

## Endpoint
`/api/deno-version` - Returns JSON with:
- Deno version details
- Window object existence
- Window.location availability

This is a temporary diagnostic tool for debugging purposes.

### DIFF
--- a/src/routes/api/deno-version/+server.ts
+++ b/src/routes/api/deno-version/+server.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { json } from '@sveltejs/kit'
 
 export const config = {


Wituareard:
isServer has nothing to do with the error, you can check the generated runtime code. And I'm pretty sure that it works correctly because it's replaced at build time by Vite and therefore doesn't depend on the environment.

anthonybailey:
It's more complicated than that, Nils. 

Paraglide makes a reasonable assumption (in many places across it's compiled runtime) that if window is defined, we are not in a server context, and that window.location.href needs inspected.

Some of them are amenable to changing by defining isServer when compiling, but turns out not all: the assumption is hard coded in other places our set up can reach, including one you've seen.

Since Netlify's production version of Deno defines a stub window object without a location property in an edge function context, we encounter the 500.

We don't really need an edge function for people/, but we would like to use one for localization redirects: it's exactly what edge functions are for, working close to the user rather than having to go through to the server or to serverless functions, both of which are less local (AWS Lambda running in different regions can help, but it isn't a CDN.)

I can see how to fix it all, I believe: it's just that quite a few places in the official paraglide release need attention.

Wituareard:
What's you plan for fixing the issues? You could contribute to the paraglide repo I suppose?

anthonybailey:
We had two options. Forking paraglide, fixing it, and depending on that is the very solid dev-correct one. It also means we could submit our patch to paraglide for their consideration, make the future better, etc.

But in order to make the near future much better, we have an easier trick: on the way into edge functions, delete the weird Netlify Deno window object whose only impact for us is to break paraglide. (We don't know why they define it. We imagine it may make some other libraries we aren't using work.)

PR #545, opened 2025-11-12
Test serverless functions

Wituareard:


Wituareard:
We could manually add an edge function that returns the geo data

Wituareard:
> What does setting edge functions to false really mean in this context, then?

The SvelteKit routes aren't served from an edge function, `/api/geo` is served from outside SvelteKit

Wituareard:
(Anthony commented on another PR that it would be good to handle l10n redirects from edge functions which makes sense)

Issue #571, opened 2025-12-02
Sidebars on Desktop


Pato-desu:
Originated here: https://github.com/PauseAI/pauseai-website/issues/433#issuecomment-3600500945
Make sidebars likes the one on mobile, but visible at the left of the page

PR #579, opened 2025-12-05
Desktop sidebars

Pato-desu:
https://github.com/PauseAI/pauseai-website/issues/571
- [ ] get feedback
- [x] make it "sticky" ? or that it scrolls showing the selected header at the middle (noticeable in risks)
- [x] fix zoomed states
- [x] check edgecases and different pages

### DIFF
--- a/src/lib/components/Toc.svelte
+++ b/src/lib/components/Toc.svelte
@@ -111,11 +111,11 @@
 
 <style>
 	:root {
-		--toc-active-color: var(--bg);
+		--toc-active-color: var(--brand);
 		--toc-li-hover-color: var(--brand);
 		--toc-li-color: var(--text);
 		--toc-mobile-bg: var(--bg);
-		--toc-active-bg: var(--brand);
+		--toc-active-bg: transparent;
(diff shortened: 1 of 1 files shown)


Wituareard:
Not sure, that would be one solution

Wituareard:
Also I think the library should already be able to scroll automatically?

Pato-desu:
> Also I think the library should already be able to scroll automatically?

Mmm, what do you mean by scrolling automatically? The sidebar in the /risks page was left showing the same headers while the highlighted headers would change to be the ones out of the screen when scrolling too much (on the corpus).

I'm basically vibe coding without understanding the code, but even activating a keepInView thing that the AI tried didn't fix it.

Pato-desu:
I thought this change was important so I asked for reviews, let me know if it wasn't necessary

Wituareard:
Yes makes sense, thanks for the reminder

Maximophone:
I don't see a sidebar when I go the the preview. Not sure what to look for. Also, I don't know what is the reason for these changes

PR #609, opened 2025-12-31
Action page: levels of engagement

Pato-desu:
Suggested by Ella.

I made a couple of small changes and converted into a table like the If you are... section.

I like it but it's a big change so I want to hear feedback.

The thing I don't like that much is the division a few hours/ a day.
It seems basically the same... and idk


Pato-desu:
- [ ] I'd change a few hours per 1 day and 1 day per 1 week, it's more slick and relatively equally inaccurate (the 1 day has a "organize events" point)

PR #616, opened 2026-01-03
Contact us page

Pato-desu:
- [x] The form is huge. too much space in between things.
- [x] Change some wording
- [x] And it is a bit overlapping with "Partnerships"
- [x] And "Press"
- [x] And even with "Send us feedback"
- [x] Not sure where the second form should send emails to, I need to create that address prob
- [x] Add mailersend integration
- [x] Have to test both forms in local 
- [x] adding API key
- [ ] Fixing whatever is happening with the API

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ deno.lock
 
 /static/open-letter/portraits/original
 CLAUDE.md
+.claude/
--- a/src/routes/contact-us/+page.server.ts
+++ b/src/routes/contact-us/+page.server.ts
@@ -4,6 +4,14 @@ import { env } from '$env/dynamic/private'
 
 export const prerender = false
 
+// Configure recipient email addresses for each contact form type
+const CONTACT_RECIPIENTS = {
+	Standard: 'info@pauseai.info',
+	Media: 'press@pauseai.info',
+	Partnerships: 'info@pauseai.info',
+	Feedback: 'info@pauseai.info'
+} as const
+
 async function sendContactEmail(data: {
 	name: string
 	email: string
@@ -20,7 +28,7 @@ async function sendContactEmail(data: {
 		}
 	}
 
-	const recipientEmail = 'patricio@pauseai.info'
+	const recipientEmail = CONTACT_RECIPIENTS[data.type]
 
 	let htmlContent = `
 		<p><strong>Name:</strong> ${data.name}</p>


Pato-desu:
@joepio You're sure everything is set up correctly? I'm having trouble on the deploy preview but it's working on local.

Gemini:
I've investigated the issue thoroughly and found the likely cause of the 500 Internal Server Error. The problem appears to be with your MailerSend configuration, which is handled in src/routes/contact-us/+page.server.ts.
Please verify the following:
MAILERSEND_API_KEY: Ensure that the MAILERSEND_API_KEY environment variable is correctly set in your Netlify project settings. This key is essential for authenticating with the MailerSend service.

anthonybailey:
If you do struggle with any "mail from Pause AI" issues, note that sending
mail under apps using Google's own SMTP at rates up to 2,000 emails/day has
proved easy, robust, and better insulated from spam detection concerns.

Just making sure people know their options.

anthonybailey:
Capturing some detail.

Joep's comment advising he changed the key is dated Jan 6 09:59 UTC and presumed after making the the change.

Pato's commit 957dc0a is marked same day PM in a not specified timezone, and appears after Joep's comment, so should have picked up the new environment value.

In Netlify said env is annotated "Same value in all deploy contexts. Joep Meindertsma updated 11 days ago" (FFS what is it with dev tools not specifying timestamps) and has prefix mlsn.416ee

The deploy preview web UX reports just "An unexpected error occurred" on submit.

anthonybailey:
Yeah here we go in edge function log:

Jan 17, 06:42:48 AM: 01KF5B2B error [render] 
[500] POST /contact-us
Error: Dynamic require of "node:https" is not supported
at file:///root/.netlify/edge-functions/render.js:15:9

The mailersend SDK isn't edge-compatible. It depends on gaxios which uses Node's node:https, and edge functions run on Deno (not the Node runtime.)

When you were running locally (perhaps under dev) you won't have manifested the same issue. You can if you use "netlify serve" but even then the version of Deno Netlify use is not directly matchable in dev.

Options:

You could move to serverless function. This would rename/move the handler so Netlify routes it through Node.js functions instead of edge. This is "correct" and conventional, but doesn't match our usual website choices so could be another can of worms for you.

Probably simplest: replace the Mailersend SDK call with a raw fetch — Mailersend's API is simple enough that you can call it directly with fetch(), which works everywhere. Eliminates the dependency.

(Really simplest as noted previously:  consistently use Google apps to send Gmail.)

@Pato-desu that should unblock you and with more help from Gemini I suspect you can determine the URL, code and test the raw fetch.

PR #618, opened 2026-01-05
removing christmas campaign

Pato-desu:


anthonybailey (commented):
Dead links are bad. Locally, I'd updated the page to close our the campaign with results, but forgot to push.

Other clean up is good.

I'll address now.

PR #625, opened 2026-01-14
Standardize Software Team volunteer role format on /join

### LINKED ISSUE #591: Add Volunteer Vacancies to the Join Page
Can we either
- add the volunteer vacancies to the join page
- have the volunteer vacancies, teams and national groups pages linked as part of the - Thank you for signing up part of the join form

Can we also link volunteer vacancies on our normal vacancies page

Anthony to update the software role in same style as the others for clarity

seankessinger:
Closes #591


anthonybailey (approved):
I'll adjust the software description a little more - think we've lost some of the useful emphases. But since bundled with other changes and blocking, approving with intent to fix after.

PR #657, opened 2026-02-20
refactor: Extract collagen signup logic from join.md

### LINKED ISSUE #529: Reduce Svelte code in join.md
Let's put it in a component instead

RisingOrange:
Closes #529

## Summary

- Extracts the collagen campaign logic (URL param detection, state management, conditional banners, newsletter pre-fill) from `src/posts/join.md` into a new `src/lib/components/CollagenSignup.svelte` component
- Exports `userHasUid`, `subscribeEmail`, and `newsletterEmail` as bound props so `join.md` can share state with the bottom newsletter signup and conditional note
- Replaces `on:click` submit detection with `on:submit` on the wrapper div — more semantically correct and doesn't rely on the synthetic click behaviour specified in the HTML implicit submission algorithm

## Note on submit tracking

The original code used a click handler to detect newsletter submission. This works because per the [HTML spec](https://html.spec.whatwg.org/multipage/form-elements.html#implicit-submission), pressing Enter in a form field fires a click event on the default submit button before firing the form's submit event. The `on:submit` approach is cleaner and more direct, but this is a code quality change rather than a bug fix.

## Test plan

- [ ] Normal visit to `/join` — page renders correctly, no banner shown
- [ ] Visit `/join?collagen_uid_sayno=xxx&subscribe-email=test@example.com` — welcome banner shown, email pre-filled in top and bottom newsletter forms, share link visible
- [ ] Submit newsletter form (click and Enter) — banner updates to "Thanks. [...]"

### DIFF
--- a/src/lib/components/CollagenSignup.svelte
+++ b/src/lib/components/CollagenSignup.svelte
@@ -11,7 +11,7 @@
 	export let userHasUid = false
 	export let subscribeEmail = ''
 	export let newsletterEmail = ''
-	let subscribeClicked = false
+	let submitted = false
 	let hideSharing = false
 
 	onMount(async () => {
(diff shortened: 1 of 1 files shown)


anthonybailey:
Sorry I took so long to notice and act. Very sensible, I should have separated it out in the first place.

PR #662, opened 2026-02-21
fix: banner pushing hero below viewport fold

### LINKED ISSUE #661: Banner pushes hero below viewport fold
## Problem

When an active banner (`<Banner>` or `<CampaignBanner>`) is displayed, the hero section on the homepage starts below the banner. Because the hero has `height: 100vh`, its bottom extends off-screen — so the \"full viewport\" hero effect is broken and users need to scroll to see the whole thing.

## Expected behaviour

The hero should fill whatever vertical space remains in the viewport after the banner(s), so its bottom edge always lands at the scroll fold.

## Affected pages

Homepage (`/`) when any banner is active — currently reproducible with `CampaignBanner` (Brussels protest).

RisingOrange:
Fixes #661

## What

When a banner is active, the hero section (`height: 100vh`) starts below it and extends off-screen, breaking the full-viewport effect.

| Before | After |
|---|---|
|  |  |

**Note:** The screenshots also show how the mobile layout is incorrectly used on vertical desktop instead of the desktop layout. This will be fixed by https://github.com/PauseAI/pauseai-website/issues/668

## How

Key changes:

- **New `page-top` wrapper** wraps banners and (on hero pages) `hero-section`. On hero pages it becomes a `height: 100dvh` flex column, so its children share the full viewport height.
- **Banners moved inside `page-top`** (previously they were in normal flow above `.layout`). They take their natural height, leaving the rest for the hero.
- **`hero-section` moved out of `.layout`** and into `page-top`, where it gets `flex: 1` to fill whatever space the banners don't use. `Hero` height changed from `100vh` to `100%` to fill its flex parent. `min-height: var(--hero-min-height)` prevents content overlap when many banners shrink the available space.
- **Shared `--page-gutter` variable on `:root`** -- the responsive padding (3rem, shrinking to 0.5rem below 600px) was previously scoped to `.layout`. Moving the hero nav out of `.layout` into full-width `page-top` meant it lost its inherited width constraint. Rather than hardcoding the padding value, the gutter is now a shared CSS variable used by both `.layout` (as padding) and the hero nav (in its width calculation: `min(var(--page-width), 100% - 2 * var(--page-gutter))`). The variables were renamed from `--padding-wide`/`--padding-narrow` to `--gutter-max`/`--gutter-min` since they now serve both padding and width purposes.
- **Flex overflow fix** -- `flex-shrink: 0` on banner elements prevents them from being squished to zero height by the hero image's intrinsic size on wide viewports (Firefox).

**Before (hero page):**
```svelte
<BannerOrNearbyEvent />           <!-- normal flow, pushes layout down -->
<CampaignBanner />                <!-- normal flow, pushes layout down -->
<div class="layout hero-page">
  <div class="hero-section">      <!-- position: relative -->
    <Hero />                      <!-- height: 100vh — overflows viewport -->
    <Header />                    <!-- nav: position: absolute over hero -->
  </div>
  <main /><Footer />
</div>
```

**After (hero page):**
```svelte
<div class="page-top hero-page">  <!-- NEW: height: 100dvh; flex column -->
  <BannerOrNearbyEvent />         <!-- takes natural height -->
  <CampaignBanner />              <!-- takes natural height -->
  <div class="hero-section">      <!-- flex: 1; min-height: var(--hero-min-height); position: relative -->
    <Hero />                      <!-- height: 100% — fills remaining space -->
    <Header />                    <!-- nav: position: absolute, width constrained via --page-gutter -->
  </div>
</div>
<div class="layout hero-page">    <!-- hero-section moved to page-top -->
  <main /><Footer />
</div>
```

## Testing

- [ ] Homepage with active banner: hero bottom at viewport fold at scroll position 0
- [ ] Dismiss banner: hero expands to fill full `100dvh`
- [ ] Resize viewport: hero adjusts without JS
- [ ] Nav width consistent between homepage and other pages at all viewport widths (especially below 600px)
- [ ] Non-hero pages: layout unchanged
- [ ] Wide viewport (2560px+) with banner: banner visible, hero fills remaining space (Firefox)
- [ ] Small viewport with many banners: hero content doesn't overlap

### DIFF
--- a/src/lib/components/Hero.svelte
+++ b/src/lib/components/Hero.svelte
@@ -116,7 +116,7 @@
 
 	.hero {
 		display: block;
-		height: 100vh;
+		height: 100%;
 		min-height: var(--hero-min-height);
 		overflow: hidden;
 		position: relative;
--- a/src/lib/components/Hero.svelte
+++ b/src/lib/components/Hero.svelte
@@ -117,7 +117,7 @@
 	.hero {
 		display: block;
 		height: 100%;
-		min-height: var(--hero-min-height);
+		min-height: min(var(--hero-min-height), 100%);
 		overflow: hidden;
 		position: relative;
 		width: 100vw;
--- a/src/lib/components/Hero.svelte
+++ b/src/lib/components/Hero.svelte
@@ -117,7 +117,6 @@
 	.hero {
 		display: block;
 		height: 100%;
-		min-height: min(var(--hero-min-height), 100%);
 		overflow: hidden;
 		position: relative;
 		width: 100vw;
--- a/src/lib/components/Hero.svelte
+++ b/src/lib/components/Hero.svelte
@@ -116,8 +116,9 @@
 
 	.hero {
 		display: block;
-		height: 100%;
-		min-height: min(var(--hero-min-height), 100%);
+		height: 100vh;
+		min-height: var(--hero-min-height);
+		max-height: 100%;
 		overflow: hidden;
 		position: relative;
 		width: 100vw;
--- a/src/lib/components/Hero.svelte
+++ b/src/lib/components/Hero.svelte
@@ -116,9 +116,8 @@
 
 	.hero {
 		display: block;
-		height: 100vh;
-		min-height: var(--hero-min-height);
-		max-height: 100%;
+		height: 100%;
+		min-height: min(var(--hero-min-height), 100%);
 		overflow: hidden;
 		position: relative;
 		width: 100vw;
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -170,9 +170,9 @@
 		top: 0;
 		left: 0;
 		right: 0;
-		/* Replicate .layout constraints: constrain to page-width minus its 3rem padding on each side,
-		   then center. This keeps --cqw container-query calculations identical to before. */
-		width: calc(min(100%, var(--page-width)) - 2 * 3rem);
+		/* Replicate the content-box width of .layout (box-sizing: content-box, max-inline-size: 40rem,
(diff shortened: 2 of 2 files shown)


RisingOrange:
@Wituareard 

> Seems to cut off the banner under Firefox for wide viewports 

Good catch! Fixed in https://github.com/PauseAI/pauseai-website/pull/662/commits/547d1fadd20ddc3ea4b57e4b12c2de531bf4cc75

src/routes/+layout.svelte
```diff
@@ -156,6 +176,8 @@
 		top: 0;
 		left: 0;
 		right: 0;
+		width: min(var(--page-width), 100vw - 6rem);
+		margin-inline: auto;
```
Wituareard: This causes inconsistencies when navigating between pages for me

src/lib/components/Hero.svelte
```diff
@@ -116,8 +116,8 @@
 
 	.hero {
 		display: block;
-		height: 100vh;
-		min-height: var(--hero-min-height);
+		height: 100%;
+		min-height: min(var(--hero-min-height), 100%);
```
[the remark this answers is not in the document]
RisingOrange: Right. Fixed in
[Set hero-section min-height to prevent content overlap on small viewp…](https://github.com/PauseAI/pauseai-website/pull/662/commits/9a6d8f0329d4b5347cbc7853c9217103891de2b1) by setting `min-height: var(--hero-min-height)` for the `.hero-section`

src/routes/+layout.svelte
RisingOrange: Fixed in [Fix nav width inconsistency between hero and non-hero pages](https://github.com/PauseAI/pauseai-website/pull/662/commits/0737a814ecdea4c8f45fa65f6b8386207aeb788d)

PR #677, opened 2026-02-28
fix: banner components close button styling

RisingOrange:
Fixes the campaign banner close button hover effect, which appeared as an oval instead of a circle because the button was stretching to the full banner height.

## Summary

- The campaign banner close button used `top: 0; bottom: 0` which stretched it to the full banner height
- With `border-radius: 50%`, this produced an oval hover background instead of a circle
- Fixed by using `top: 50%; transform: translateY(-50%)` to give the button a fixed size and proper vertical centering

## How to test

The campaign banner is currently disabled in `+layout.svelte`. To reproduce, temporarily restore [this version of the condition](https://github.com/RisingOrange/pauseai-website/blob/78f8d1abcf7df936cf431819e5ce584019d16e23/src/routes/%2Blayout.svelte#L92), and hover over the close button on a multiline banner.

## Before / After

| Before | After |
|--------|-------|
| Before | After |

### DIFF
--- a/src/lib/components/Banner.svelte
+++ b/src/lib/components/Banner.svelte
@@ -129,9 +129,9 @@
 
 	.close {
 		position: absolute;
-		top: 0;
+		top: 50%;
 		right: 0.75em;
-		bottom: 0;
+		transform: translateY(-50%);
 		display: flex;
 		align-items: center;
 		background: transparent;


Wituareard:
Thanks, I applied the same fix to the regular banner component

PR #701, opened 2026-03-26
PauseAI Australia pages additional content and formatting tweaks

### LINKED ISSUE #697: Take out GDoc link from pauseai.info/australia
In the _**Get involved**_ section, the final dot point is a link to a Google Doc. This provides many suggestions for additional actions, but can be confusing for new people. We should only offer it as an option for volunteers that did the basic tasks. The dot point should be taken out.

I am happy to have this assigned to me, or to have anyone else action.

PeterHorniak:
PauseAI Australia pages additional content and formatting tweaks
We are also testing using svelte components from the base site, using a script.


Wituareard:
You may want to crop and/or reduce the three images, but I'll merge this for now, and it can be done later

PR #706, opened 2026-03-26
Fix geo country cookie TypeError

RisingOrange:
`geo.country` from Netlify's edge function context is an object `{ code?: string; name?: string }`, not a string. This caused two bugs introduced in #695:

- **TypeError in production**: Passing the object as `SetCookie`'s `value` caused `quote()` to call `.includes()` on an object → `TypeError: t.includes is not a function`
- **Cookie reset on every page load**: Comparing the object to the string cookie value always evaluated to `true`, so the cookie was re-set and `selectBanners()` was called on every page visit

## Fix
Use `geo.country.code` (the string country code) instead of `geo.country` in both the comparison and the cookie value.

## Test plan

The geo API returns 501 in dev (no Netlify platform), so these can only be verified on a deploy preview:

- [ ] No `TypeError: t.includes is not a function` in the console
- [ ] The `geo_country` cookie is set to a country code string (e.g. `"US"`), not `[object Object]`

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -36,7 +36,7 @@
 	onMount(async () => {
 		const response = await fetch('/api/geo')
 		if (!response.ok) return
-		const geo = await response.json()
+		const geo: GeoApiResponse = await response.json()
 
 		// Keep geo cookie in sync with actual location.
 		// Re-run selectBanners if country changed or cookie not yet set.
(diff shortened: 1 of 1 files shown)


Wituareard:
Thinking about whether we should check for `[object Object]` or maybe broader "invalid country code" for the people who visited in the meantime

PR #723, opened 2026-04-03
Press 3 cards for row with images

Pato-desu:


Wituareard:
You should delete the pnpm-lock.yaml and re-install to get rid of the 3k lines diff

PR #747, opened 2026-04-09
Add Skeleton component for latest news section

Wituareard:
This should be fairly uncontroversial, but I'll wait a bit with merging it


Wituareard:
Actually I think it's uncontroversial enough and the improvement big enough to merge it immediately

PR #762, opened 2026-04-12
Add statement condemning attack on Sam Altman's residence

Maximophone:


Wituareard:
"the attacker" (without "suspected" or anything) seems inconsistent with "the suspect"

Maximophone:
> "the attacker" (without "suspected" or anything) seems inconsistent with "the suspect"

good catch. Trying to fix it with the CMS but not sure I did the right thing

PR #775, opened 2026-04-18
Restructure /funding to be a Markdown post

### LINKED ISSUE #756: Restructure /funding to be Markdown post
Having the content spread across Markdown files makes it harder to edit and maintain.

The rationale for the current approach is that it completely renders on the server during pre-rendering via a load function. We could achieve the same thing using the [new experimental await expression](https://svelte.dev/docs/svelte/await-expressions) or have the client call an API, maybe assisted by the new Skeleton component

RisingOrange:
Closes #756. Converts `/funding` from a custom route into a standard Markdown post under `src/posts/`, matching every other content page — while keeping the donor list prerendered at build time.

Also fixes a latent bug: the previous split `intro.md`/`afterDonors.md` lived under `src/routes/funding/`, so they were outside the `src/posts` → `l10n-cage/md/<locale>/<slug>.md` pipeline and always fell back to English regardless of locale. Now they're translated like every other post.

## Summary

- New `src/posts/funding.md` rendered by the generic `[slug]` route; deletes the custom `src/routes/funding/` directory (`+page.svelte`, `+page.server.ts`, `+page.ts`, `meta.ts`, `intro.md`, `afterDonors.md`).
- Donor list fetched via a SvelteKit `prerender` remote function (`src/lib/funding-donors.remote.ts`) and consumed from a small child component (`FundingDonorsList.svelte`) via an `{#await}` template block — keeps the list baked into the static HTML, no client-side flash.
- Removes the now-redundant `fundingMeta` entry from `src/routes/api/posts/+server.ts` (the post is auto-discovered via the `src/posts/*.md` glob).

## Tradeoff: experimental flag

Enables one experimental flag in `svelte.config.js`:
- `kit.experimental.remoteFunctions = true` (SvelteKit remote functions)

**Why this over the alternatives:**
- A client-side fetch with a loading state would lose prerendering (slower, worse at scale, and a visible flash).
- Adding a slug-specific hook inside the generic `[slug]/+page.server.ts` would leak funding-specific logic into a route meant to be dumb.
- The flag is additive — nothing else in the repo needs changes, and the blast radius of a future API change is one component.

## Test plan

- [x] `pnpm check` — 0 errors
- [x] `pnpm build` — `/funding` prerenders; static HTML contains the donor list
- [x] Donor list rendering verified locally with mock data; the Notion fetch itself is unchanged from the previous route
- [ ] Reviewer verifies production build with real Notion credentials still populates the list


Wituareard:
For some reason this causes the whole page to briefly disappear and re-appear in both Chrome and Firefox

RisingOrange:
@Wituareard 

Right, I couldn't reproduce it locally, but I saw the flash on the preview deploy.

This should be fixed by [Use {#await} block instead of top-level await for donor list](https://github.com/PauseAI/pauseai-website/pull/775/commits/15b410ba11307af8fcdfb7c7d4b46baef1c404f1)

RisingOrange:
Reproduced it locally with `pnpm build && pnpm preview` and confirmed the fix

Wituareard:
Yeah if those are the options then having the list flash is better than having the whole page flash I guess

Wituareard:
I really hoped that experimental await without a boundary would behave like load functions tbh

RisingOrange:
The list flash should be brief, because the data is still prerendered (no live Notion call).
We could check on the deploy preview.

Either way, probably still worth it?

Wituareard:
The list definitely isn't prerendered in the preview, but the data might be? Either way I agree that it's probably still worth it, thanks!

PR #778, opened 2026-04-19
Restructure /pdoom to be a Markdown post

RisingOrange:
Same motivation as #775, #776 — convert a custom route into a standard Markdown post so it flows through the normal content pipeline.

`/pdoom` is the cleanest candidate yet: 30 lines of pure prose wrapped around an embedded `<Doomers />` component. No server data, so **no experimental flags** needed.

## Summary

- New `src/posts/pdoom.md` replaces `src/routes/pdoom/` (`+page.svelte`, `meta.ts`).
- `<Doomers />` component already lives in `$lib/components/` and is imported directly in the markdown's `<script>` block.
- Removes the `pdoomMeta` entry from `src/routes/api/posts/+server.ts`.

## Visual change

The title (and everything below it) shifts up slightly — the `[slug]` layout has a bit less top spacing above the heading than the old custom route did. This matches how every other Markdown post on the site positions its title; aligning with the rest of the content pipeline is the point of the refactor.

<table><tr>
<td align="center"><strong>Before</strong><br />Before</td>
<td align="center"><strong>After</strong><br />After</td>
</tr></table>

## Test plan

- [x] `pnpm check` — 0 errors
- [x] `pnpm build` — `/pdoom` prerenders; built HTML contains the Doomers list


Wituareard:
Ugh there seems to be an issue with components in Markdown where Svelte doesn't link the CSS in the HTML, so the component isn't styled when loading, probably still worth it

RisingOrange:
@Wituareard 
> Ugh there seems to be an issue with components in Markdown where Svelte doesn't link the CSS in the HTML, so the component isn't styled when loading, probably still worth it

Looking into it

Wituareard:
Thanks, two possible causes I found:
- mdsvex after vitePreprocess
- Markdown components imported dynamically in `[slug]/+server.ts`

RisingOrange:
> Markdown components imported dynamically in [slug]/+server.ts

It's this one. 

The issue is also visible on https://pauseai.info/donate. The Donate button jumps around on page reload. 

I think I'm on to a good solution

PR #779, opened 2026-04-19
Fix whitespace above title on short pages (SearchModal grid stretch)

RisingOrange:
Moves `<SearchModal>` out of `header.svelte` to the root layout so `<pagefind-modal>` stops occupying a row in the `.layout` grid.

## Before / After

/outcomes at viewport 1440×1800 — currently live on prod:

<table><tr>
<td>Before: empty gap on /outcomes</td>
<td>After: no gap on /outcomes</td>
</tr></table>

Other prod pages also affected at tall viewports: /vacancies (~630px gap), /funding (~88px gap).

## The bug

`.layout` in `+layout.svelte` uses `grid-template-rows: auto 1fr auto`. `<SearchModal>` was rendered as a sibling of `<Navbar>` inside `header.svelte`, so `<pagefind-modal>` ended up as a direct grid child of `.layout`:

```
┌──────────────────┐   grid rows
│ navbar           │   auto
├──────────────────┤
│ pagefind-modal   │   1fr   ← absorbs empty space
│   (empty unless  │
│    opened)       │
├──────────────────┤
│ main             │   auto (implicit)
├──────────────────┤
│ footer           │   auto
└──────────────────┘
```

On pages whose content is shorter than the viewport, pagefind-modal stretched to fill the `1fr` row.

## The fix

Render `<SearchModal>` from `+layout.svelte` as a sibling of the layout grid (it's a global overlay, not header content). The open state moves to `src/lib/stores/searchModal.ts` so the header's search button can still trigger it.

## Test plan

- [ ] Visit /outcomes, /vacancies, /funding, or any short page on a tall viewport — title sits directly under the nav, no empty gap.
- [ ] Click the search icon in the nav — modal opens as before.

### DIFF
--- a/src/lib/stores/searchModal.ts
+++ b/src/lib/stores/searchModal.ts
@@ -1,0 +1,3 @@
+import { writable } from 'svelte/store'
+
+export const searchOpen = writable(false)
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -7,7 +7,9 @@
 	import Link from '$lib/components/Link.svelte'
 	import NearbyEvent from '$lib/components/NearbyEvent.svelte'
 	import PreloadFonts from '$lib/components/PreloadFonts.svelte'
+	import SearchModal from '$lib/components/SearchModal.svelte'
 	import Toc from '$lib/components/Toc.svelte'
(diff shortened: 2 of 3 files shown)


Wituareard:
Good catch, thanks

PR #780, opened 2026-04-19
Resolve Svelte hydration warning caused by ESLint comment in <svelte:head>

Wituareard:
Fixes a TypeError: e.getAttribute is not a function hydration warning in the console.

In Svelte 5, placing an  comment directly inside a <svelte:head> block causes the comment to be rendered as an HTML node. During hydration, Svelte expects an Element node and tries to call .getAttribute() on it, which throws the error.

Moving the eslint-disable and eslint-enable comments to the outside of the <svelte:head> block resolves the hydration mismatch while keeping the linter happy.


Wituareard:
Okay nvm this doesn't resolve it

PR #787, opened 2026-04-20
Inject per-post CSS links to eliminate FOUC on Markdown posts

RisingOrange:
Fixes FOUC for Svelte components used inside `src/posts/*.md`. Spotted on #778 (thanks @Wituareard for flagging the root cause).

## The bug

Markdown posts are loaded via a dynamic `import()` in `src/routes/[slug]/+page.ts`. That import boundary means any component imported inside a `.md` file lives in its own async chunk — so its CSS file isn't in the `[slug]` route's CSS manifest, and isn't linked in the prerendered HTML. The browser renders the post markup with unstyled components, then hydration kicks in and Vite injects the `<link>` at runtime, causing a visible flash / layout shift. (Visible for example on `/donate` — the donate button jumps from left-aligned to centered on load.)

## The fix

- `+page.server.ts` reads the Vite client manifest (`.svelte-kit/output/client/.vite/manifest.json`) at SSR time, walks the transitive `imports` graph starting from the post's source path, and returns the list of CSS asset URLs.
- `+page.svelte` injects a `<link rel="stylesheet">` for each via `<svelte:head>`.
- `+page.ts` spreads the server load's data through so the URLs reach the component.

Each post ends up linking exactly the CSS it needs — no manual component list to maintain, no bundle bloat.

## Test plan

- [x] `pnpm build && pnpm preview`, open `/donate` — no button jump.
- [x] Compared before/after build output: e.g. `donate.html` now links `donate.css` + `mdsvex.css`; `join.html` adds `NewsletterSignup.css` + `TallyEmbed.css`.
- [x] `pnpm check` passes.
- [x] Lint: no new errors; pre-existing warnings only.

## Notes / caveats

- Only affects prerendered HTML (which is how all posts render today). If any slug ever moves to SSR-on-demand, the Netlify function bundle would need to include the manifest file — worth a follow-up if that comes up.
- Dev mode has no manifest; the helper returns `[]` (dev doesn't have FOUC anyway since Vite injects styles via JS).

### DIFF
--- a/src/routes/[slug]/+page.server.ts
+++ b/src/routes/[slug]/+page.server.ts
@@ -3,7 +3,5 @@ import type { PageServerLoad } from './$types'
 import { cssForPost } from './post-css.server'
 
 export const load: PageServerLoad = ({ params: { slug } }) => {
-	const locale = getLocale()
-	const src = locale === 'en' ? `src/posts/${slug}.md` : `l10n-cage/md/${locale}/${slug}.md`
-	return { cssUrls: cssForPost(src) }
+	return { cssUrls: cssForPost(slug, getLocale()) }
 }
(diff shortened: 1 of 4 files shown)


Wituareard:
That's a pretty cool solution ngl, thank you!

PR #832, opened 2026-05-02
Testing Tally webform with black background

PeterHorniak:
Updated TallyEmbed component to include height property and added custom styles.


Wituareard:
approved the preview

Wituareard:
Changing the styling without previewing locally probably doesn't make sense so maybe you should either put the content changes on a new branch or revert the styling changes here

PR #835, opened 2026-05-03
Update syntax to Svelte 5

mzdunek93:
There are still some `svelte/legacy` imports that should be removed, but that can be finished later.


anthonybailey:
Linking https://svelte.dev/docs/svelte/v5-migration-guide as a courtesy. Michael you might describe any interesting choices you made relative to that?

mzdunek93:
@anthonybailey The vast majority of the modifications were done by the migration script, and consists of prop migrations (declaring the interface type and destructuring the `$props` variable), rewriting variables into using `$state` or `$derived`, etc. Sometimes I compressed a variable declaration and modification in effect to a single `$derived` declaration, like in `Banner.svelte` with `dismissed`. Sometimes I had to initialize a variable that was not initialized, like with `type` in `Link.svelte`, since otherwise there would be an [error](https://svelte.dev/docs/svelte/bind#bind:property-for-components) with passing `undefined` to a `$bindable()` with fallback. `LinkWithoutIcon.svelte` changed a bit, since I converted the `smoothScroll` action into an attachment. In order to avoid `svelte/legacy` imports, I changed event handlers to invoke `event.preventDefault()` or `event.stopPropagation()` in places like `chat/+page.svelte`. In some places like `CommunitiesList.svelte`, I wasn't sure if a `const` is supposed to be reactive, but I made it `$derived`, since otherwise I got a warning. I also added some API response type declarations in `uk-mp-contact-status.ts‎` and `uk-postcode-to-mp.ts` in order to suppress some eslint warnings, which is not really relevant to the purpose of this PR, but it let me pass the checks.

mzdunek93:
@anthonybailey Also in `TallyEmbed.svelte` I used `$derived.by` for `urlParams` instead of `$effect` to make sure the variable is properly initialized in SSR.

Wituareard:
We could create a svelte5 branch with the result of the script to get a diff with the manual adjustments (I personally can't atm) but not sure if that's necessary

mzdunek93:
@Wituareard Good idea. I ran the migration script on a recent version of main and uploaded it without changing anything. [Here's the diff](https://github.com/mzdunek93/pauseai/compare/31c6b51..6e6b1a4) with the current version of this branch. It's much shorter (+338/-364 instead of +1203/-740) so should be easier to read.

Wituareard (commented):
(abort review)

Wituareard (commented):
(abort review again, I just want to comment)

Wituareard:
Thanks! Are you sure about `$derived` in `Banner.svelte`? Iiuc it makes `dismissed = true` temporary, the same goes for `hasImageError` in `NewsCard.svelte`. The rest looks good to me, but I have yet to do proper testing

mzdunek93:
@Wituareard You are right about `Banner.svelte`, I misinterpreted how it's supposed to work. In `NewsCard.svelte`, that variable is supposed to become `false` every time an image link is passed, but then might be changed to `true` if `NetlifyImage` fails. So it's supposed to be temporary. But I reversed the value of the variable by mistake. The current diff is [here](https://github.com/mzdunek93/pauseai/compare/06c36de..3907f4).

Wituareard:
Makes sense, I'll check it sometime in the next few days

Wituareard:
I looked at ` hasImageError` again and decided to replace the slightly unintuitive `$derived` solution with an `$effect` one

Wituareard:
Image.svelte didn't correctly resolve paths on the server because the `$effect` block only ran in the browser

Wituareard:
- The styling of `Toc.svelte` is off in some ways, though the upgrade fixes the backdrop which is currently broken in production
- I'm getting a bunch of [`non_reactive_update`](https://svelte.dev/docs/svelte/compiler-warnings#non_reactive_update) warnings in the build log
- The build fails with
```
Error: The following routes were marked as prerenderable, but were not prerendered because they were not found while crawling your app:
 - /[slug]
 ```

mzdunek93:
@Wituareard build should work fine now (also changed the svelte/legacy import to a slightly different method and remove more effects). The current diff is [here](https://github.com/mzdunek93/pauseai/compare/e8b5a6b..5656d75).

Wituareard:
@mzdunek93 Do you plan to address the build warnings?

mzdunek93:
@Wituareard fixed svelte warnings

Wituareard:
Thanks! The metadata type already exists in types.ts, could you use that one?

mzdunek93:
@Wituareard fixed.

Wituareard:
The mobile ToC is broken, I can only open it once, and it doesn't have a backdrop

Wituareard:
Maybe we should do this in multiple steps
- Update svelte-toc and Toc.svelte only
- Run migration steps + strictly necessary adjustments only (+ comment placement)
- Remove legacy imports etc

Wituareard:
Looks good now, thank you for doing this and also thanks for your patience

PR #866, opened 2026-05-14
[DIAG] Surface Airtable error in MP-email response

UFO-101:
Temporary diagnostic patch. The UK MP email endpoint is currently returning 500 in production with an opaque message; this surfaces the underlying Airtable error text in the response body so we can identify the cause without function-log access.

Test via the Netlify deploy preview's `/uk-email-mp` form — submit, then read the response body in the network tab. **Revert immediately after.**


Wituareard:
Can we close this?

UFO-101:
Oh yeah, sorry, we don’t need this now.

PR #871, opened 2026-05-16
Update svelte-toc and migrate Toc.svelte to runes

Wituareard:
The font size of the desktop layout is still slightly too large. ~~I also had to sacrifice the enlarged hitbox of the close button.~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,11 +29,6 @@
 		"prepare": "lefthook install",
 		"sync": "svelte-kit sync"
 	},
-	"pnpm": {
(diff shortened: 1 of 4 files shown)


Wituareard:
It's not pixel-perfect but close enough I think

Wituareard:
The backdrop works locally but not in the preview

Wituareard:
Since the backdrop is broken anyway I'll just push to main to see if the Netlify JS is interfering

PR #872, opened 2026-05-16
Migrate Backdrop.svelte to runes

Wituareard:


Wituareard:
Still doesn't work, whatever

PR #876, opened 2026-05-18
Update most dependencies

mzdunek93:
The only dependencies which haven't been updated here are the API ones, which will be done later.


Wituareard:
thanks

PR #877, opened 2026-05-18
Add Parliament-visit checkbox to UK MP email form

UFO-101:
Adds a checkbox above the message field for users planning to attend [PauseAI UK's June 23rd visit to Parliament](https://luma.com/q2wu0y59) to speak with their MP in person.

When checked, the message's "Next steps" block is swapped for a bold sentence proposing an in-person meeting at Parliament. If the user has already customised the Next steps section, the sentence is inserted before the signature instead, so their edits aren't lost. Unchecking restores the message exactly as it was before the box was checked.

The luma link carries `utm_source=uk-email-builder`.

Styled as a soft brand-tinted card with a custom checkbox (the native one was being stretched to 100% by the global `input` rule and pushing the label text below it). The card background intensifies when ticked.

### DIFF
--- a/src/lib/components/UKMPEmailForm.svelte
+++ b/src/lib/components/UKMPEmailForm.svelte
@@ -45,6 +45,36 @@ ${userPostcode.toUpperCase()}`)
 	let submitStatus: 'idle' | 'success' | 'error' = $state('idle')
 	let errorMessage = $state('')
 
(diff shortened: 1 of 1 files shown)


UFO-101:
Merging this as it shouldn't have any effects outside of the UK email builder.

PR #914, opened 2026-06-11
Optimize news card image sizes

Wituareard:


Wituareard:
The latest version of enhanced-img only generates one resolution for ?enhanced imports it seems

Wituareard:
We should monitor the next few builds to see if caching works or if this adds 5 minutes to each build

Wituareard:
A manual deploy I triggered took the usual amount of time so all good it seems

Issue #897, opened 2026-05-31
Load libraries for homepage asynchronously


Wituareard:
The homepage currently loads (tree-shaked parts of) a few heavy libraries that don't affect content above the fold (Glide, NumberFlow, Motion). They should probably be loaded asynchronously to speed up initial load times. At least we should measure their impact.

- [x] Glide
- [x] NumberFlow
- [x] Motion

PR #915, opened 2026-06-11
Load Motion lazily

Wituareard:
See #897


Wituareard:
The bundle definitely gets smaller, not sure why Lighthouse is worse

Wituareard:
Just noticed that it might be the JS that Netlify injects

Wituareard:
Contributions/investigation welcome

Wituareard:
Took another look at the chunk sizes and decided to merge and see how it performs

PR #951, opened 2026-07-01
feat: add ?city and ?languages prefill params to onboarding embed

hturnbull93:
## Summary

- Adds `?city=<value>` query param to prefill the city field in the onboarding embed iframe
- Adds `?languages=<comma-separated>` query param to prefill the languages multiselect (e.g. `?languages=German,English`)
- Invalid/unrecognised language values are silently dropped; falls back to `['English']` if nothing valid is provided
- Consistent with the existing `?country=` param behaviour

## Test plan

- [ ] `?city=Berlin` → city field prefilled with "Berlin"
- [ ] `?languages=German,English` → languages multiselect prefilled with German and English
- [ ] `?languages=InvalidLang` → falls back to English (invalid value dropped)
- [ ] `?country=Germany&city=Berlin&languages=German` → all three params work together
- [ ] No params → city empty, languages defaults to English (existing behaviour unchanged)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015aKhZeP6x8pgBkDEraXGUs


src/lib/components/onboarding/OnboardingFlow.svelte
anthonybailey, line 1: Implementation of logic for defaulting an empty set of *valid* languages to just Englush looks wrong? Check size after filter, no? L122-124.

anthonybailey, line 1: I see the existing pattern being extended with more specific cases, but seems possibly better to generalize and simplify. Why not allow an initial value for any field?

anthonybailey, line 1: How are we currently documenting or tracking who embeds and what we support? External interfaces need curation.

hturnbull93: sure, documented now

hturnbull93: the fields `country`, `city` and `languages` are the only fields that can sensibly be prefilled, the rest of the fields would be particular to an individual

hturnbull93: well spotted, fixed :)

hturnbull93:
@anthonybailey happy to rereview?

Wituareard:
Could you move the Markdown doc to the new docs directory? The rest looks good

PR #958, opened 2026-07-06
Document onboarding flow [skip netlify]

Wituareard:
@hturnbull93 Could you verify that the flow is depicted correctly if you still remember it roughly?


hturnbull93 (approved):
Yes, more or less correct, was this generated from looking at the code as it is at the moment?

Wituareard:
> was this generated from looking at the code as it is at the moment?

yes

Wituareard:
I'll probably just remove the sequence diagram and go ahead since the flowchart is what motivated this

PR #702, opened 2026-03-26
Improve post banner framing and social preview compatibility

RisingOrange:
Closes #659

Use a consistent `1200 / 628` (~`1.91:1`) crop for post banners. This ratio looks good on the website and is also the recommended one for social media previews. Add a `showImage: false` frontmatter option for posts where the image should not be used as a banner, and switch a few posts to cropped image variants so the wider banner is framed better and avoids awkward cutoffs.

## Summary

- Use `aspect-ratio: 1200 / 628` with `object-fit: cover` for banner images instead of `max-height: 400px`
- Add `showImage: false` frontmatter option to disable the banner while keeping the image for OG/social previews
- Set `showImage: false` on `/india-summit-2026` and `/4-levels-of-ai-regulation`, and place cropped versions of the images inline in the article body in a constrained-width wrapper instead
- Use adjusted image assets for the problematic square/near-square posts (`india-summit-logo.png`, `4levels-cropped.png`)
- Switch the Brussels, Amsterdam, and Stop the AI Race posts to cropped photo variants so the wider banner is framed better and avoids awkward cutoffs like heads getting cropped:
  - `/protests-san-francisco-cropped.jpeg`
  - `/banner-all-cropped.jpeg`
  - `/stop-the-ai-race-march-cropped.jpg`

## Recommendations for banner images going forward

**Use landscape images around `1200×628px` if possible.** That ratio works for both the website banner and the common Open Graph preview shape used by Twitter/X, Facebook, Slack, Discord, etc.

If the image gets cropped badly in the website's `1.91:1` banner, adjust it so it looks good at that ratio on the website. Then it should also look good in link previews (though some platforms can still crop or render cards a bit differently). For photos where the wider crop cuts off heads or frames the scene awkwardly, use separate cropped source files so the original photos can still be preserved in the repo.

If you want the image somewhere other than the top of the article, set `showImage: false` in frontmatter and place it inline in the article body instead.

This can also be useful when the available image does not work well as a wide banner. In those cases, the image can still be kept for social/OG previews if there is no better landscape alternative yet, since a non-ideal preview image is still better than none.

### `/india-summit-2026`
<table><tr>
<th>Before</th><th>After</th>
</tr><tr>
<td>india-summit before</td>
<td>india-summit after</td>
</tr></table>

<p><em>Banner disabled; cropped image moved inline at constrained width.</em></p>

### `/4-levels-of-ai-regulation`
<table><tr>
<th>Before</th><th>After</th>
</tr><tr>
[...]

PR #987, opened 2026-07-18
Improve news card image framing and text alignment

RisingOrange:
Two related improvements to the news cards: images are no longer cropped, and card text lines up across a row.

## Image framing

The cards used `aspect-ratio: 16/10` with `object-fit: cover` for every image, cropping the edges of anything that wasn't already a matching landscape shape — posters, charts and Substack banners lost content.

Following the existing post-banner design system (#702), which standardizes internal post images to a curated `1200/628` crop:

- **Card frame** now uses `aspect-ratio: 1200 / 628` to match the post banner ratio.
- **Internal post images** keep `object-fit: cover` — they're curated to frame well at this ratio. Their `Image` `aspectRatio` hint is updated to `1200/628` so the responsive `sizes` math stays accurate.
- **Substack/press images** (arbitrary shapes we don't control) use `object-fit: contain`, so charts, posters and tall portraits show in full instead of being cropped.
- The image container gets a `--bg-subtle` background so the contained letterbox blends with the card body.
- Hover zoom (`scale(1.03)`) is scoped to cover-cropped images only — zooming a contained image would re-crop it under `overflow: hidden`.

## Text alignment

Titles run from one to three lines, so each card's description started at a different height and short titles left a gap above their description.

- The title now reserves a **fixed three-line area** and is **vertically centered** within it, so descriptions line up regardless of title length. Three lines (not two) because several real titles need three at the site's condensed heading font — two would clip them mid-word.
- The 2-line clamp lives on an inner span, because `-webkit-line-clamp` needs `display: -webkit-box`, which can't coexist with the flex centering. The title area is a **column** flex so the loading skeleton bars still stack.
- Dates already align (pinned to the bottom of equal-height cards).

## Before / After

**These are real screenshots of the running site** (dev server, live news data, actual fonts and theme) — not mockups.

In *before*: PauseCon's wordmark is clipped, Fable 5 is cut off on both sides, the chart and portraits are cropped, and the row-1 descriptions ("Apply now" / "Hundreds of…" / "The US reaction…") sit at three different heights. In *after*: images show in full and the descriptions line up.

**Before:**

Before

**After:**

After

## Verification
- `pnpm check` — 0 errors; `pnpm _lint:knip` — passes; eslint clean
- Rendered and captured from the running dev server with live prod news data, confirming `object-fit: contain` resolves correctly for external images inside NetlifyImage's `<picture>` wrapper, that no title is truncated, and that the loading skeleton still stacks.


Wituareard:
The screenshots looks very different from what I get in production (font etc), why is that?

hturnbull93:
probably better to add a landscape image rather than portrait ones probaby? On the vertical centering of the title and the subtitle I have no preference, I thought the originals looked fine but I am easy to please when it comes to website cards

Wituareard:
The image framing sounds like a good idea but the background should be different, maybe the light brand color on dark and vice versa? Or maybe a blurred/tinted copy of the image? Not a big fan of the text alignment tbh, it's not clear what the description aligns to so it looks a bit unintentional.

PR #1010, opened 2026-07-29
Show the critical-alert notice on the signup form

RisingOrange:
Shows the critical-alert notice on the signup form. It was written into the form but commented out, so no signup has ever seen it. Maxime asked for it to be shown.

## Why the checkbox doesn't already cover it

The case is someone who ticks **only the required privacy-policy box** and neither optional box. They are still added to the critical alerts list in CiviCRM, but nothing in the form tells them any email may follow. The required box covers the privacy policy and chapter sharing, and the policy's only email disclosures are the opt-in Substack newsletter and "contact you via Airtable" for coordination.

## Wording

> We may occasionally send you a critical alert, even if you don't opt into any of these. See our [privacy policy].

The critical-alert notice under the two opt-in checkboxes on the intent step

"any of these" points at the two checkboxes above it. "critical" and "occasionally" keep this a bounded tier rather than "we may email you anything", which would undercut the opt-in it bypasses. "critical alert" matches the unsubscribe page name (**PauseAI Critical Alerts**) — in English, at least; that page is English-only.

## Implementation

- **Step 2, under the two opt-in checkboxes**, not the email field where the commented-out draft sat: step 1 shows no opt-ins for the clause to point at, and submits nothing (`onsubmit={continueToIntent}`), so no data leaves the browser before step 2.
- **Not on the browse mini-form**, which hardcodes `keep_informed=on` — submitting it *is* the opt-in, so the clause would be false there. Its required privacy-policy checkbox still carries the general disclosure.
- **`aria-describedby` on both controls**, so the notice is not a paragraph a screen-reader user tabs straight past. First use in this file: the other helper texts are hints, this one is a disclosure.
- **`onboarding_email_critical_notice` in `messages.ts` (en/de/fr)** rather than inline English, following `onboarding_gdpr_consent`, which also carries the privacy link as inline markup via `{@html}`.

## Checks

CI green: `pnpm-check`, CodeQL, and a Netlify deploy preview worth clicking through to see it on a real build.

The **de and fr wording is mine and wants a native-speaker read** — four defects surfaced in it during review, which no check here would have caught.

Context: PauseAI/pauseai-civicrm#389 and PauseAI/pauseai-civicrm#344.

### DIFF
--- a/src/lib/components/onboarding/OnboardingFlow.svelte
+++ b/src/lib/components/onboarding/OnboardingFlow.svelte
@@ -422,10 +422,8 @@
 						autocomplete="email"
 						bind:value={basics.email}
 					/>
-					<!-- <p class="helper">
-						We may contact you about critical mobilizations, see our
-						<Link href="/privacy">privacy policy</Link>.
-					</p> -->
+					<!-- eslint-disable-next-line svelte/no-at-html-tags -->
+					<p class="helper">{@html msgs.onboarding_email_critical_notice}</p>
 				</div>
 				<div class="field">
 					<label class="field-label" for="ob-country">{msgs.onboarding_field_country}</label>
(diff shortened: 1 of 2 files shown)


hturnbull93 (approved):
Lgtm, though best to show Maxime and Irina to make sure this is ok GDPR wise. I think there's a chance we may need to put it by the submit button as a mandatory checkbox, as we might not be able to have the submit itself assume consent.

RisingOrange:
Folding this into #1014, which carries this notice as part of the reworked step 2 (same mechanics — the aria-describedby wiring and message keys — plus a list-unsubscribe reassurance in the copy). The de/fr native-speaker caveat moves there with it.

PR #1023, opened 2026-08-05
Extract /about loader component and convert route to markdown post

Wituareard:
## Summary

Converts the `/about` SvelteKit route into a markdown post (`src/posts/about.md`), served by the existing `[slug]` route. The data-fetching, loading state, and rendering logic that previously lived in the route's `+page.svelte` is extracted into a reusable loader component.

## Why

Markdown is generally easier to maintain than a SvelteKit route — content edits don't require touching component logic, and the file stays focused on prose. It also integrates with our Git-based CMS workflow. Markdown posts are also auto-discovered by the `[slug]` route via `import.meta.glob`, so they don't need to be registered in `hardCodedPages`. Extracting the people-loading logic into a component makes it reusable and keeps the markdown focused on content.

## Changes

- **New** `src/lib/components/about/AboutPeople.svelte` — loader component handling the `/api/about` fetch, loading skeletons, and grouped rendering of people.
- **Moved** `src/routes/about/person.svelte` → `src/lib/components/about/PersonCard.svelte` (unchanged content).
- **New** `src/posts/about.md` — the `/about` content as a markdown post, importing `AboutPeople` and `Link` via mdsvex.
- **Deleted** `src/routes/about/+page.svelte` and `src/routes/about/meta.ts` (route no longer needed).
- **Modified** `src/routes/api/posts/+server.ts` — removed the `aboutMeta` import and its entry in `hardCodedPages`, since the post is now auto-discovered.

## Notes

- Redirects pointing at `/about` (e.g. `/people`, `/about-us`) continue to work via the `[slug]` route.
- The `/api/about` endpoint is unchanged; `AboutPeople` imports its `AboutApiResponse` type from there.

## Known regressions

Some custom styling from the original route was not carried over to the markdown post:

- The first headings ("How We Began", "Key Milestones", "Non-violence") were pretty much identical to the default `.prose` styling, so they render the same.
- The **"Essential Information"** heading looks slightly different — it previously had a larger font size and a top border separating it from the section above.
- The **essential info list** no longer spans two columns on wider screens; it now renders as a single-column list (the original used a 2-column grid at `min-width: 768px`).


Wituareard:
**Tradeoff: people list is no longer server-rendered / indexable.** Previously (before #1022), `+page.svelte` fetched the people data server-side via `load`, so it was present in the initial SSR'd HTML and indexable by crawlers. Since this content is now a markdown post rendered via mdsvex, there's no `load` function available to it, so `AboutPeople` has to fetch `/api/about` client-side after mount instead. That means the people list is absent from the server-rendered output and only appears once the fetch resolves in the browser — crawlers that don't execute JS (and some that do but don't wait for async client-side fetches) won't see it.

Wituareard:
Follow-up in #1024: it adds a server-side data-loading path for markdown posts via a frontmatter `data` map, so posts like `/about` can fetch URL-backed data during SSR and include it in the initial HTML — restoring server-rendering/indexability without giving up the markdown-post workflow.

Wituareard:
Our Netlify setup has [(legacy) prerendering](https://docs.netlify.com/build/post-processing/prerendering/#legacy-netlify-prerendering) enabled so we don't need to worry about indexability

PR #1024, opened 2026-08-06
Allow markdown posts to opt out of prerendering and load URL-backed data

Wituareard:
## Summary

Markdown posts can now opt out of prerendering and fetch URL-backed data at request time via a frontmatter `data` map.

Resolves the tradeoff raised in #1023 (comment): markdown posts had no `load` function available, so components like `AboutPeople` had to fetch data client-side after mount, leaving the people list absent from server-rendered HTML and unindexable by crawlers. This PR gives markdown posts a server-side data-loading path via frontmatter `data`, so that content can be fetched during SSR and included in the initial HTML.

## Changes

- **Allow markdown posts to opt out of prerendering** — posts can set `prerender: false` in frontmatter to be dynamically rendered instead of baked into static HTML at build time.
- **Load URL-backed data in markdown posts via frontmatter `data` map** — a new `data` frontmatter field maps names to URLs that are fetched server-side and passed to the mdsvex component as a `data` prop. JSON responses are parsed; other responses are passed through as text. Documented in `README.md`.
- **Cache dynamically rendered posts for 1 hour** — non-prerendered posts now set `Cache-Control: public, max-age=3600` so CDN/edge caches can serve repeated requests without hitting the origin.

## Usage

```markdown
---
title: Live Example
description: Example post with request-time data.
prerender: false
data:
  news: /api/news?page=1&pageSize=3
---

{#each data.news.items as item}
<h2>{item.title}</h2>
{/each}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,42 @@ The article (and automatic l10ns of same) will be previewable.
 
 If you are sufficiently changing prominent text, consider inspecting relevant l10ns as well as the original.
 
+### Loading Data in Markdown Posts
+
+Markdown posts can fetch URL-backed data on the server by adding a `data` map to frontmatter. This is most useful for internal API routes:
+
+```markdown
+---
+title: Example
+description: Example post
+data:
+  news: /api/news?page=1&pageSize=3
+---
+
+<script lang="ts">
+	export let data
+</script>
+
+{#each data.news.items as item}
+<h2>{item.title}</h2>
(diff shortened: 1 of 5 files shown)


Wituareard:
Our Netlify setup has [(legacy) prerendering](https://docs.netlify.com/build/post-processing/prerendering/#legacy-netlify-prerendering) enabled so we don't need to worry about indexability

PR #1014, opened 2026-08-01
Untangle Step 2 of the signup form (opt-ins vs intent)

RisingOrange:
**For reviewers:** the model in the "What each choice results in" table is decided (2026-08-20 meeting with Irina), so it needs no sign-off — it documents what each choice leads to. What wants review eyes is the implementation itself. The de/fr strings were aligned with the vocabulary of the national chapters' own pages (pause-ai.de, pauseia.fr): both say Gruppe / groupe local, so the earlier Kapitel/chapitre calques are gone file-wide. To try the form live: [open the preview](https://deploy-preview-1014--pauseai.netlify.app/embed/onboarding-form) (the reworked step is page 2).

<table>
<tr><td align="center"><b>Before — form on main</b></td><td align="center"><b>After — this PR</b></td></tr>
<tr>
<td valign="top">Before — current Step 2</td>
<td valign="top">After — reworked Step 2</td>
</tr>
</table>

## What it changes (Step 2)

Mostly it *untangles* the current intent step rather than redesigning it. Two different kinds of choice are jumbled together there today:

- **Opt-ins** — do you want our emails? The 🔔 "Keep me informed" and 📰 Substack checkboxes.
- **Intent** — how active do you want to be? Act now / Volunteer / Lead.

The proposal cleans up the specifics:

- **🔔 copy** — "Get global campaign updates, plus news and ways to help from your local chapter," replacing "connect me with my local chapter", which wrongly implied chapter contact was *gated* by the checkbox (it isn't — a volunteer who skips it is still reachable about their volunteering via legitimate interest). "and ways to help" is deliberate: with one audience per chapter, a bell-ticker receives the chapter's local calls to action too, so copy promising news alone would understate what they get.
- **📰 Substack copy** — "AI news from the PauseAI team's perspective," replacing "general news on AI, delivered via our Substack newsletter." The two boxes now read as clearly different things (🔔 = what PauseAI does; 📰 = news about AI, our take), rather than two vaguely-newsletter-ish checkboxes side by side.
- **The newsletter stays a genuine opt-in for everyone**, including Volunteers/Leads — nothing pre-checked or forced (marketing consent must be freely given).
- **Critical-alert floor notice** under the opt-ins (copy from #1010, plus a list-unsubscribe reassurance).
- **No "tick at least one" requirement**, so a critical-alerts-only sign-up is reachable — the state the notice's premise assumes.
[...]

PR #1031, opened 2026-08-09
Newsletter subscribe entry points + dedicated /subscribe signup

RisingOrange:
Adds newsletter subscribe entry points and a dedicated `/subscribe` signup, so people who just want updates can find and complete a signup without going through the full "Join" flow.

Requested and discussed in [this Discord thread](https://discord.com/channels/1100491867675709580/1222902730083799120/1535226046293479425).

> [!NOTE]
> **The Airtable side is done.** The Members-table automations now read the local-chapter choice this form adds:
>
> - **"Onboarder assignment + alert (new member)"** only hands someone to their national chapter's leader if they agreed to be connected with a local chapter. Anyone who did not is routed to the global onboarding address instead, which keeps their verification email working, since it is CC'd to whoever is assigned.
> - **"Share US signups with PAI US"** only copies a US signup into the shared Google Sheet if they agreed to that too.
> - A new automation notifies the chapter when one of these subscribers later chooses to volunteer or lead, since that choice escalates the sharing anyway.
>
> Signups through the existing Join form are unaffected: that form's privacy checkbox already covers chapter sharing, so every one of its rows carries the agreement. The broader intent/consent question in #1014 is untouched by this.

### Entry points

Subscribe item in the header nav
<p align="center"><em>A "Subscribe" item in the header</em></p>

Homepage subscribe box
<p align="center"><em>The existing signup box near the bottom of the homepage — it now carries the email into the form instead of subscribing straight to Substack</em></p>

### The signup

Subscribe form
<p align="center"><em>/subscribe</em></p>

Thanks + do-more prompt
<p align="center"><em>After signing up</em></p>

### "Do more" → Get involved

Continues in place, reusing the /join flow without creating a duplicate:

Get involved intent step
<p align="center"><em>Choosing how to get involved</em></p>

Volunteer form under Get involved
<p align="center"><em>e.g. the volunteer path</em></p>

## What this changes

- **Nav:** a "Subscribe" entry (mail icon) in the header, before Donate. Adding a sixth item pushed the logo onto its own line on every page, so the header now gets a little more width than the text column it used to share. That part is a site-wide change, worth a look on a few pages.
- **Homepage box:** the existing newsletter box near the bottom of the homepage now leads into the signup form (carrying the email) instead of subscribing straight to Substack, and its heading reads "Subscribe".
- **`/subscribe`:** a dedicated single-page newsletter signup — name, email, country, city, with two optional opt-ins:
  - "Also send me updates from my local chapter"
  - "Also subscribe me to the PauseAI Substack"
  - Name, email, country and city are all required (as on the /join form). Privacy consent is the act of signing up (linked microcopy, no checkbox).
- **"Do more" continuation:** after signing up, an optional step to get more involved (Act now / Volunteer / Lead). It continues on the same page rather than sending people to `/join`, and updates the signup they just made instead of creating a second one.

  This step is the `/join` intent step, minus its two email checkboxes ("Keep me informed" and the Substack). On `/join` that is where those choices get made; here they were just made on the form a moment earlier, so repeating them would only invite someone to switch off the newsletter they came for. The choices they made are kept as they are.

## Consent behaviour

- Global newsletter is the default (signing up is the opt-in).
- Chapter sharing happens **only** when the local-chapter box is ticked. This decouples chapter sharing from the privacy consent on the subscribe path; the existing `/join` form is unchanged (it still bundles the two).
- Choosing Volunteer or Lead does turn chapter updates on, since both mean working with a local group. If someone picks one of those and then backs out again, the choice they made on the form is restored.

## Notes for reviewers

- Live writes are gated behind `ONBOARDING_LIVE`; the Netlify preview runs in **stub mode**, so submitting the form writes nothing to Airtable or Substack — feel free to click through.
- The subscribe form's copy is currently English-only (hardcoded), unlike the neighbouring onboarding components that route through the localization layer — worth deciding whether to localize it.

_Related: signup Step 2 proposal #1014. The CRM side of the local-chapter choice is done (PauseAI/pauseai-civicrm#493) — a decline recorded here now reaches CiviCRM, keyed on this form's own signup source._


hturnbull93:
We now require name email country and city on Substack signup, so a bit confused why we need to change the airtable automations? Surely we capture everything needed for them?

hturnbull93:
Unless we mean that if someone subscribes in a country without a national leader, then they get assigned to a global handler?

RisingOrange:
@hturnbull93 Right now the **Onboarder assignment + alert** automation notifies an onboarder of someone's country about every new signup, and it doesn't know about the new "Also send me updates from my local chapter" checkbox. Same for **Share US signups with PAI US**. And picking Volunteer or Lead in the get-involved step only updates the row, so nothing fires and the onboarder never hears about it.

I think I have the updates we should do to the automations nearly worked out.

hturnbull93:
> @hturnbull93 Right now the **Onboarder assignment + alert** automation notifies an onboarder of someone's country about every new signup, and it doesn't know about the new "Also send me updates from my local chapter" checkbox. Same for **Share US signups with PAI US**. And picking Volunteer or Lead in the get-involved step only updates the row, so nothing fires and the onboarder never hears about it.
> 
> I think I have the updates we should do to the automations nearly worked out.

Ok cool, I think those templates got adjusted to include the action/intent they chose, right?
On the "lead" intent they are prompted to email Irina so less critical the national lead gets told about it. Also I think we have checks for if there is already an appropriate existing leader? I can't recall if I put that in here or for catalyse.

hturnbull93 (approved):
Lgtm, happy with the discussion and can see the new flow in the screenshots

RisingOrange:
> I think those templates got adjusted to include the action/intent they chose, right?

Yeah, the intent is shown in the mail.

> On the "lead" intent they are prompted to email Irina so less critical the national lead gets told about it.

Good point, this means that the gap only really matters for volunteers.

>  Also I think we have checks for if there is already an appropriate existing leader?

Yes, the existing-leader check is in there: it looks up a chapter for their country and falls back to Irina + global onboarding when there isn't one. We can reuse this branch for people who leave the local-chapter box unchecked, so they get the global onboarder instead of a national one

Wituareard:
@RisingOrange Could you maybe update `docs/join-form-flow.md`?

RisingOrange:
> @RisingOrange Could you maybe update `docs/join-form-flow.md`?

Good point, updated it in https://github.com/PauseAI/pauseai-website/pull/1049

PR #1036, opened 2026-08-12
Add Turnstile verification to the contact forms

Maximophone:
## Why

The `/contact-us` forms' only anti-spam measure was a hidden honeypot field ([#646](https://github.com/PauseAI/pauseai-website/pull/646)). That catches naive form-fillers, but does nothing about bots that POST directly to the form actions — a single `curl` with an `Origin` header reached partnerships@, press@ or feedback@ and also triggered an auto-reply to an attacker-supplied address, putting our MailerSend sending reputation at risk.

## What

- **Cloudflare Turnstile, verified server-side** before any email is sent. Fails closed if `TURNSTILE_SECRET_KEY` is missing, except under `dev`. Uses SvelteKit's compile-time `dev` rather than `isDev()`, so the check doesn't depend on which env vars the runtime exposes.
- **Deleted the `standard` action.** Its UI tab was removed in 0600d92d, so nothing legitimate could reach it — anything arriving there was a bot.
- **Escaped user input in the notification email.** It was interpolated raw, letting spammers inject live links into the team's inbox.
- **Server-side validation** of lengths, the 200-word message cap, email format, and the partnership-type select. These were browser-only hints before.

The honeypot stays — it's free and catches a different class of bot.

## Setup required

Two Netlify environment variables:

| Name | Notes |
|---|---|
| `PUBLIC_TURNSTILE_SITE_KEY` | Public. **Do not** mark as secret — it appears in build output and secrets scanning would fail the deploy. Read at build time, so it needs the Builds scope. |
| `TURNSTILE_SECRET_KEY` | Secret. Needs the Functions/Edge Functions scope. |

Leave `TURNSTILE_SECRET_KEY` unset for Deploy previews and Branch deploys so preview URLs fail closed rather than acting as an unprotected relay. `template.env` documents Cloudflare's test keys for local development.

## Verification

Tested against the dev server with Cloudflare's dummy keys:

| Case | Result |
|---|---|
| Direct POST with no token, all 3 tabs | 403, blocked |
| Always-fail secret with a well-formed token | 403, blocked |
| Valid token | passes gate, reaches MailerSend |
| Honeypot filled | silent fake success |
| `partnership_type` not in our list | 400 |
| Message >200 words / name >100 chars / malformed email | 400 |
| `?/standard` | 404 |

In the browser: the token enables submit, a failing key leaves submit disabled with an explanation, and tab switching leaves exactly one widget with no duplicates.

**Note:** `vite preview` doesn't surface `$env/dynamic/private` in this project, so the production-build path with a working secret couldn't be tested locally. The mechanism is the same one `MAILERSEND_API_KEY` already uses in production. Worth making one real submission after deploy to confirm.

## Limits

Turnstile stops automated submissions, not a human manually typing spam. If that starts, per-IP rate limiting would be next — awkward here because edge functions are stateless.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/app.d.ts
+++ b/src/app.d.ts
@@ -1,4 +1,5 @@
 import type { Locale } from '$lib/paraglide/runtime.js'
+import type { TurnstileApi } from '$lib/turnstile'
 import type { Context } from '@netlify/edge-functions'
 import type { Picture } from 'vite-imagetools'
 // See https://kit.svelte.dev/docs/types#app
@@ -29,6 +30,8 @@ declare global {
 		selectBanners(): void
 		applyTheme(): void
 		dataLayer?: unknown[]
+		turnstile?: TurnstileApi
+		onPauseAITurnstileLoad?: () => void
 	}
 
 	declare module '*.md' {
--- a/src/lib/components/Turnstile.svelte
+++ b/src/lib/components/Turnstile.svelte
@@ -1,0 +1,87 @@
+<script lang="ts">
+	import { onMount } from 'svelte'
+	import {
+		loadTurnstile,
+		TURNSTILE_FIELD,
+		turnstileSiteKey,
+		type TurnstileApi
+	} from '$lib/turnstile'
+
+	// Tokens are single-use and expire after five minutes, so the parent remounts
+	// this component (via a {#key} block) after every submission.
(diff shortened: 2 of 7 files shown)


Maximophone:
Expediting merge because we are under attack by a bot

PR #1058, opened 2026-08-21
style: format 2024-february.md, unbreaking CI on main

RisingOrange:
`src/posts/2024-february.md` picked up a double blank line in ae9345ff, which prettier rejects, so `pnpm check` is failing on `main` and every open PR inherits it through the merge ref.

The fix is `prettier --write` on that one file: a single blank line removed, no content change.

Worth knowing for next time: edits made through GitHub's own web editor bypass the local pre-commit formatter, so they can land unformatted even though the same edit made locally could not. (The CMS does run Prettier, per @Wituareard below, so this is specific to the GitHub UI.)


Wituareard:
The CMS does run Prettier but GitHub UI doesn't

RisingOrange:
@Wituareard Thanks, corrected it

Wituareard:
Thanks, btw we also have the "Manual Format" workflow for this kind of thing (or /format on PRs)

PR #1103, opened 2026-09-11
Onboarding email follow-ups: match the live copy, fixed lines, per-chapter overrides

RisingOrange:
Follow-ups to #1065, one commit per topic so anything you disagree with can be dropped on its own. `pnpm check`, `pnpm lint` and the tests pass.

The base branch `feat/onboarding-email-render-endpoint` is a copy of #1065's head (cde8f5d1), pushed here so this PR shows only these commits and gets a deploy preview. Merging it there means these changes land inside #1065.

**To read the emails**, open [/onboarding-email-preview](https://feat-onboarding-email-followups--pauseai.netlify.app/onboarding-email-preview) on this branch's deploy: every combination of intent, country and language, in HTML or plain text. [/onboarding-email-compare](https://feat-onboarding-email-followups--pauseai.netlify.app/onboarding-email-compare) puts a render next to the MailerSend template it replaces.

## What changed

**The copy matches what is live.** The generic path sent one long email to every intent. Since Irina's confirmation copy went live, no intent and Act now get a short email and only Volunteer and Lead get the long one, so there are three versions rather than four, ported from what the live templates send. `Keep informed` is written by no form any more and falls to no intent.

**Every email is now greeting, confirm link, body, newsletter line, sign-off.** The confirm link and the newsletter and critical alert line come from `fixed.ts` in the email's language, so neither the shared copy nor a chapter's own email can leave them out — the UK's was missing both. Inside an email a chapter wrote, that promise takes a shorter form: ours reads as boilerplate in somebody else's note.

**Chapters keep the emails they wrote.** PauseAI UK keeps both of its current emails rather than one for every intent, PauseAI Canada keeps its volunteer email, which this PR would otherwise have dropped to the shared copy, and PauseAI Sverige has a new one, in the words their lead already sends by hand. The two lines the skeleton adds around those words are our Swedish, which nobody fluent has read yet: worth less than it sounds against the alternative, which is those readers getting the English email. An override supplies the subject, greeting, body and sign-off, and takes its links from the chapter's Airtable row wherever a field holds them, so link rot has one place to be fixed.

**The chapter block promises contact only to volunteers**, since picking Volunteer or Lead includes consent to share with the chapter. Non-volunteers get the chapter's links with no promise, and nothing when the row has none. It names the chapter rather than the leader, because for several chapters the onboarding alert goes to someone else or to a role address. Spanish-speaking countries get no chapter block, so Spain is not singled out from the rest of PauseAI en Español, and PauseAI US onboards its own members, so US signups get none either.

**It is closer to the email it replaces, though not a copy of it**: 16px body text, a 26px title with the name in the brand orange, the PauseAI icon above it, the video as a button, the verification sentence between rules, and the sender's accounts as icons under the footer rule. Those icons are the ones the templates use, rehosted here rather than hotlinked from MailerLite's CDN. The markup is ours rather than MailerLite's, so the two do not match line for line, and the compare page is the place to judge how close is close enough.

**The Airtable script is the version that runs.** The copy in #1065 could not have worked: `setTimeout` does not exist in Airtable's sandbox, so the render call would have rejected immediately and every send would have taken the fallback, and that fallback routed non-volunteers to a retired template.

**The unsubscribe endpoint and line are gone.** MailerSend appends its own footer, as a domain setting, to raw sends as well as template sends, so the email would have carried two links. The endpoint also mutated on a bare GET and suppressed the address across the whole domain.

**Robustness**: a failed chapter lookup renders without a chapter rather than failing the send, chapter links are escaped and limited to web links, a signup's own name can no longer render as markdown in an email that is also CC'd to their chapter's onboarder, and the chapter fetch asks Airtable for the fields it reads rather than every column of every row.

**The QA pages are built for a copy review**: re-render in place, a frame the height of the email, a country picker that says which chapters replace the shared copy, and colours that survive a browser forcing its own dark scheme. The compare page's pickers no longer pair a non-volunteer template with the volunteer email.

French is not in this PR. Nothing routes to it today, since the Canada FR template has never sent, and French for France should wait until that chapter has read it.

## Needs copy review, but not before merging

Each email here reproduces what that audience gets today, apart from Sweden's, which replaces an English email with their chapter's own words. So these can follow the release.

- **Joseph**: both UK emails, and one question. The UK volunteer template runs a MailerSend split test, so UK volunteers currently get one of two emails at random: the long styled one, or the same short note non-volunteers get. A composed send has to pick one, and this reproduces the long one for volunteers and the note for non-volunteers.
- **PauseAI Canada**: the volunteer email. Its export is confirmed current, so the only change is structural: the newsletter sentence is now the shared fixed line, which adds the critical alert promise.
- **A fluent Spanish reader**: one new sentence, the critical alert promise at the end of the newsletter line.
- **PauseAI Sverige**: the two lines we add in Swedish, the confirm sentence and the newsletter line. Their own words need no review; ours do.


hturnbull93:
We may not want to do logos for the social media links in case an email client blocks images. it's less important if it's our logo png which doesn't really do anything for the user, but the social media link logos become unusable
image

RisingOrange:
Fair point, though I'd keep the icons. The templates we're replacing already use 18px icons with alt text, so it isn't new, and the footer row is the least functional part of the email: confirming the address, the chapter's links and the next steps are all text links in the body, so a reader with images off loses nothing they need.

I'd be fine with going back to text links though if you want.

I pushed a fix for the icon fallback: our markup pinned each icon to `height: 18px` with `display: block`, which clipped the alt text into an 18px box when images are blocked. The old templates set only the width, so the label can take the space it needs. I've matched that, and capitalised the labels, so a blocked footer reads as words rather than boxes:

The email footer with images loaded and with images blocked

The labels stay clickable when they fall back, since the link wraps the image rather than the other way round.

hturnbull93:
ok, I understand there are copy review things that can be done post launch, so unless there's anything else, let's go

RisingOrange:
@hturnbull93 Fine with me! Merged this one into your PR branch

PR #1116, opened 2026-09-14
Add Dutch and US-China Pause Button pages

joepio:
Hey folks,

Two updated proposal pages. One for China-US, one for NL (ASML).

## What

Two new pages on how a pause on superintelligence could actually be enforced, plus links from Building the Pause Button.

- **`/asml-pause-button` The Dutch Pause Button.** The Netherlands already licenses every export of ASML's EUV machines, and since 2024 also servicing, spare parts and software updates. The page proposes changing the licensing *criteria* rather than the instrument: a layered set of measures (chip registration, capacity cap, fab-level KYC, inspection access, then flow-down to chip buyers), a four-phase proposal, the legal basis under Article 9 of EU Regulation 2021/821 and its gaps, objections (FDPR retaliation, China catching up, cost to ASML), and open questions.
- **`/us-china-pause-button` The US-China Pause Button.** Where Washington and Beijing stand ten days before Xi's 24 September visit, why the talks stall (both sides describe a verification problem), what the treaty literature proposes (MIRI, Belfield, our own compute governance report, Convergence Analysis) versus what the DC think tanks advise, and why a supplier state can provide the verification layer neither superpower can offer the other.


anthonybailey (changes requested):
Both new pages seem excellent to me. No notes. Very current, very useful. Let's promote them.

The old page is the only way to the new pages at the moment. It thinks GPT4 is a frontier model.

By all means keep and publish a historic copy of the page as-is too, but it needs a complete refresh. You can be as merciless as you like: nuance can stay on the old page on the first pass and just disappear from this one. We want enough context to understand the ASML + China/US proposal, nothing more is needed.

(We can provide other ways to the new pages too, of course. The current page is already somewhat buried.)

anthonybailey (approved):
Great! Addressed.

PR #1141, opened 2026-09-18
Add the three 2026 protests to the archive

jacquesallen:
The list on /protests currently stops at 13 December 2025, so three actions from this year are missing, even though write-ups for all three already live in this repo.

- 23 February, Brussels, outside the European Parliament (`src/posts/brussels-ep-protest-2026.md`)
- 28 February, London (`src/posts/protest-london-feb-2026.md`)
- 21 March, San Francisco (`src/posts/stop-the-ai-race-march.md`), organised by People for a Pause with PauseAI taking part

Only `src/posts/protests.md` changes: a new `### 2026` heading with three entries, newest first, matching the existing format and linking to posts that are already published. The San Francisco line carries a parenthetical for the organiser, in the same style as the existing "(World Pause Coalition)" and "(not officially PauseAI)" notes.

I could not find a PauseAI protest between April and August 2026, so if something happened in that window it still needs adding. Worth a check against your own records rather than my search results.

Written with Claude and reviewed by me before opening.


Wituareard:
Thanks! "Organized by People for a Pause" doesn't seem accurate though, is this deliberate?

